PR #26, opened 2024-09-27
Basic Police rank functionality, police page generation, and correct …

NerdyNinja11:
…colouration of police names


AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+            )
+        ]
+
+        self.ranks = GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, DEFAULT_RANKS)
+
+    def ask_set_ranks(self):  # TODO Make Ctrl-C work - not sure why it isn't
+        return [
+            Label("Rename different ranks of police to fit the game theme"),
+            Label("Ranks affect position on the police page, and can be used to reward/punish/just have a bit of fun"),
+            InputWithDropDown(
```
jsyiek, line 100: How about a [LargeTextEntry](https://github.com/jsyiek/AU2/blob/master/AU2/html_components/LargeTextEntry.py)? You could have each rank delimited by a newline. This would make bulk modification easy (as opposed to modding one at a time), and would allow you to have endless ranks.

This would mean you'd use a list and not a dictionary to deal with ranks. Your default ranks could then by `[Constable, Chief of Police, Umpire]` with the latter two hard-coded never to be reached by players. The rank up would then be a relative offset (so ask them how many ranks up or down to offset a police).

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+    def answer_set_ranks(self, htmlResponse):
+        rank = htmlResponse[self.html_ids['Rank']]
+        new_name = htmlResponse[self.html_ids['Rank Name']]
+        if new_name:
+            GENERIC_STATE_DATABASE.arb_state[self.identifier][rank[0]] = new_name
+            return [Label(f"[POLICE] Successfully renamed rank {rank} to {new_name}")]
+        else:
+            return [Label(f"[POLICE] Rename failed - no name given")]
+
+    def on_event_request_create(self) -> List[HTMLComponent]:  # TODO Make a dependent selector, ideally with police filtering
```
jsyiek, line 120: nit: put the TODO on its own line within the function (and elsewhere)

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+                ]
+            )
+        ]
+
+    def on_event_update(self, e: Event, htmlResponse) -> List[HTMLComponent]:
+        e.pluginState.setdefault(self.identifier, {})["rank"] = htmlResponse[self.html_ids["Rank"]]
+        return [Label("[POLICE] Success!")]
+
+    def on_page_generate(self, htmlResponse) -> List[HTMLComponent]:
+        pass
```
jsyiek, line 159:
```suggestion
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+        alive_police: List[Assassin] = [i for i in police if i not in dead_police]
+        if GENERIC_STATE_DATABASE.plugin_map.get("WantedPlugin", True):  # i.e. we can separate police into dead and corrupt-dead
+            pass  # TODO after wanted rewrite
+            '''
+            dead_corrupt_police: List[Assassin] = [i for i in dead_police if i.]
+            dead_non_corrupt_police: List[Assassin]
+            '''
+
+        tables = []
+        # Going to assume there are police, because otherwise you wouldn't have this plugin enabled
```
jsyiek, line 181: I would not make this assumption in any way that could cause a crash. It plays badly with the (e.g.) `RandomGame` plugin, and could play poorly with games like May Week depending on how AU2 gets generalized.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+        if alive_police:
+            alive_police.sort(key=lambda a: (-int(police_rank_manager.get_rank(a)), a.real_name))
+            rows = []
+            print(GENERIC_STATE_DATABASE.arb_state.get(self.identifier))
```
jsyiek, line 185:
```suggestion
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+                        EMAIL=a.email,
+                        COLLEGE=a.college,
+                        NOTES=a.notes
+                    )
+                )
+            tables.append(
+                POLICE_TABLE_TEMPLATE.format(ROWS="".join(rows))
+            )
+            if dead_police:
+                dead_police.sort(key=lambda a: (-int(police_rank_manager.get_rank(a)), a.real_name))
```
jsyiek, line 201: nit: I would sort with priority given to `a.college` then `a.real_name`.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+                        identifier=self.html_ids["Rank"],
+                        title="Promote police? Must be a number from 0-9",
+                        default=e.pluginState.get(self.identifier, {}).get("rank", {}),
+                    )
+                ]
+            )
+        ]
+
+    def on_event_update(self, e: Event, htmlResponse) -> List[HTMLComponent]:
+        e.pluginState.setdefault(self.identifier, {})["rank"] = htmlResponse[self.html_ids["Rank"]]
```
jsyiek, line 155: "Must be a number from 0-9" but no validation

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+                title="Enter new name",
+                identifier=self.html_ids["Rank Name"],
+            )
+        ]
+
+    def answer_set_ranks(self, htmlResponse):
+        rank = htmlResponse[self.html_ids['Rank']]
+        new_name = htmlResponse[self.html_ids['Rank Name']]
+        if new_name:
+            GENERIC_STATE_DATABASE.arb_state[self.identifier][rank[0]] = new_name
```
jsyiek, line 115: You need to play evasively whenever setting into a dictionary (as you've done elsewhere), e.g.:

```suggestion
            GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {})[rank[0]] = new_name
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+                self.answer_set_ranks
+            )
+        ]
+
+        self.ranks = GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, DEFAULT_RANKS)
+
+    def ask_set_ranks(self):  # TODO Make Ctrl-C work - not sure why it isn't
+        return [
+            Label("Rename different ranks of police to fit the game theme"),
+            Label("Ranks affect position on the police page, and can be used to reward/punish/just have a bit of fun"),
```
jsyiek, line 99: I think umpires will know what's going on if they're using this plugin, so it should be safe to delete

```suggestion
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+        self.config_exports = [
+            ConfigExport(
+                "PolicePlugin_set_ranks",
+                "Police -> Set Ranks",
+                self.ask_set_ranks,
+                self.answer_set_ranks
+            )
+        ]
+
+        self.ranks = GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, DEFAULT_RANKS)
```
jsyiek, line 94: I think you intend to avoid having to `.setdefault` throughout code with this, but I'm not sure it will work as you intend. If the user nukes the database and then enables the `PolicePlugin`, the `__init__` doesn't get re-run and `GENERIC_STATE_DATABASE["PolicePlugin"]` won't exist (`__init__` is only run on program start)

AU2/plugins/util/PoliceRankManager.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Simple manager for police ranks
+    """
+
+    def __init__(self):
+        # from assassin ID to deadline
+        self.assassin_ranks = defaultdict(lambda: DEFAULT_POLICE_RANK)
+        self.activated = GENERIC_STATE_DATABASE.plugin_map.get("PolicePlugin", False)
+
+    def add_event(self, e: Event):
```
jsyiek, line 34: I think you should stick with the standard of +1 kill = +1 rank to avoid the umpires having to remember. Of those umpires I spoke to, police rank-ups were the most unanimously desired feature to be automated that wasn't already automated in AU1.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+    def on_event_create(self, e: Event, htmlResponse) -> List[HTMLComponent]:
+        e.pluginState.setdefault(self.identifier, {})["rank"] = htmlResponse[self.html_ids["Rank"]]
+        return [Label("[POLICE] Success!")]
+
+    def on_event_request_update(self, e: Event) -> List[HTMLComponent]:  # TODO Make a dependent selector, ideally with police filtering
+        return [
+            Dependency(
+                dependentOn="CorePlugin_assassin_pseudonym",
+                htmlComponents=[
+                    AssassinDependentTextEntry(
```
jsyiek, line 144: [AssassinDependentIntegerEntry](https://github.com/jsyiek/AU2/blob/master/AU2/html_components/AssassinDependentIntegerEntry.py)

jsyiek: I really should just have an AssassinDependentEntry factory that takes one of a few pre-defined validators to avoid having to make a new AssassinDependentEntry component for different validators, but oh well. I'll refactor it if I ever have time/if it's ever worth it.

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -55,7 +55,7 @@
 ]
 
 
-INCO_COLORS = [
+INCO_COLS = [
```
jsyiek, line 58: good spot and clean-up

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+    def __init__(self):
+        super().__init__("PolicePlugin")
+
+        self.html_ids = {
+            "Rank": self.identifier + "_rank",
+            "Rank Name": self.identifier + "_name"
+        }
+
+        self.config_exports = [
+            ConfigExport(
```
jsyiek, line 86: could you add a config option to disable the police manual rank-up field?

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -0,0 +1,228 @@
...
+from typing import List
+
+from AU2 import ROOT_DIR
+from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
+from AU2.database.EventsDatabase import EVENTS_DATABASE
+from AU2.database.GenericStateDatabase import GENERIC_STATE_DATABASE
+from AU2.database.model import Assassin, Event
+from AU2.html_components import HTMLComponent
+from AU2.html_components.AssassinDependentIntegerEntry import AssassinDependentIntegerEntry
+from AU2.html_components.AssassinDependentSelector import AssassinDependentSelector
```
jsyiek, line 13: This import doesn't appear needed. Can you check which of these you need and delete those you don't?

jsyiek (changes requested):
Great start, a few things to consider

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: Would ranking down (to below default) work then, and if so, how would the LargeTextEntry recognise the default?

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: I don't see college being relevent at all for police, so name seems to make more sense (also why I've removed police addresses from the website)

AU2/plugins/custom_plugins/PolicePlugin.py
jsyiek: One possible solution is to make the "default default" the lowest rank list. And then if users want to override that, allow them to append `=default` to the end of the ranks. You can deal with it when parsing the output of `LargeTextEntry`.

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: Personally I had no clue what police ranks were for in my game, but I guess this would make more sense in the docs/umpiring guide

AU2/plugins/custom_plugins/PolicePlugin.py
jsyiek: makes sense

AU2/plugins/custom_plugins/PolicePlugin.py
jsyiek: if you think it adds value you can leave them. But I imagine leading with a `Label` is what's causing your back-out failure. Let me check out to your branch and confirm.

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: I did encounter this a few times while testing; I assume it can be fixed by using this setdefault everywhere the database is read (in which case do I even need it in __init__()?)

jsyiek: It's harmless in `__init__`. I'd leave it there. But yes, the first access to the state in any function must use `.setdefault`.

AU2/plugins/custom_plugins/PolicePlugin.py
jsyiek: Why you can't go back is nothing you've done... I'll get this sorted in another branch.

jsyiek: https://github.com/jsyiek/AU2/pull/27

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
                 "Police -> Set Ranks",
                 self.ask_set_ranks,
                 self.answer_set_ranks
+            ),
+            ConfigExport(
+                "PolicePlugin_toggle_auto",
```
jsyiek, line 98: Not crucial, but as you mentioned you're concerned about config bloat, you could probably merge these into one config option with a [`SelectorList`](https://github.com/jsyiek/AU2/blob/master/AU2/html_components/SelectorList.py) (all you're doing is toggling on/off, so selection with the list = true, and not = false).

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
+            ConfigExport(
+                "PolicePlugin_toggle_police_kills_rankup",
+                "Police -> Toggle Ranking up on Police Kills",
+                lambda *args: [],
+                self.answer_toggle_police_kills_rankup
             )
         ]
 
-        self.ranks = GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, DEFAULT_RANKS)
+    # I've done the database stuff in such a way that configs are only stored if changed; otherwise the hardcoded default configs are used.
```
jsyiek, line 117: This is sensible, and is good design. It saves us file space on our generally limited SRCF provision!

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
+    # Not sure how sensible this is, but it seems to work
+
+    def gsdb_get(self, plugin_state_id):
+        return GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get(self.plugin_state[plugin_state_id]['id'], self.plugin_state[plugin_state_id]['default'])
```
jsyiek, line 121: nit: Can you try line-wrapping this (you can end a line with a backslash to avoid breaking it, e.g. something like this (but don't commit exactly this suggestion without testing)

```suggestion
    def gsdb_get(self, plugin_state_id):
        return GENERIC_STATE_DATABASE.arb_state   \
                        .get(self.identifier, {}) \
                        .get(self.plugin_state[plugin_state_id]['id'], self.plugin_state[plugin_state_id]['default'])
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
 
-    def ask_set_ranks(self):  # TODO Make Ctrl-C work - not sure why it isn't
+    def gsdb_set(self, plugin_state_id, data):
+        GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {})[self.plugin_state[plugin_state_id]['id']] = data
+
+    def ask_set_ranks(self):
+        # TODO Make Ctrl-C work - not sure why it isn't --> Check Jamie's pull
```
jsyiek, line 127: nit

```suggestion
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
+        default_text = self.gsdb_get("Ranks").copy()
+        default_text[int(self.gsdb_get("Default Rank"))] += "=default"
```
jsyiek, line 129: very nit: You might consider checking that your default rank index does exist, and resorting to some sensible default otherwise (to avoid crashing in the event of an umpire fiddling with the DB).

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
-        rank = htmlResponse[self.html_ids['Rank']]
-        new_name = htmlResponse[self.html_ids['Rank Name']]
-        if new_name:
-            GENERIC_STATE_DATABASE.arb_state[self.identifier][rank[0]] = new_name
-            return [Label(f"[POLICE] Successfully renamed rank {rank} to {new_name}")]
-        else:
-            return [Label(f"[POLICE] Rename failed - no name given")]
+        answer = htmlResponse[self.html_ids['Ranks']].split("\n")
+        if answer[0] != "Add named ranks, rename old/autogenerated ranks, and set default rank." or answer[1] != "The final and penultimate ranks are reserved for Umpire and Chief of Police":
+            return [Label(f"[POLICE] Rename failed - Instructions tampered with")]
```
jsyiek, line 143: You could consider starting the instructions with `#`, and deleting all lines that start with `#`. That way, you don't need to fail the user after they've given input - just leave as is if they mess with `#`. You could explain to them that lines starting with `#` are ignored (in another instr line).

```python
answer = [a for a in answer if not a.lstrip().startswith("#")]
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
     def answer_set_ranks(self, htmlResponse):
-        rank = htmlResponse[self.html_ids['Rank']]
-        new_name = htmlResponse[self.html_ids['Rank Name']]
-        if new_name:
-            GENERIC_STATE_DATABASE.arb_state[self.identifier][rank[0]] = new_name
-            return [Label(f"[POLICE] Successfully renamed rank {rank} to {new_name}")]
-        else:
-            return [Label(f"[POLICE] Rename failed - no name given")]
+        answer = htmlResponse[self.html_ids['Ranks']].split("\n")
+        if answer[0] != "Add named ranks, rename old/autogenerated ranks, and set default rank." or answer[1] != "The final and penultimate ranks are reserved for Umpire and Chief of Police":
```
jsyiek, line 142: If you're going with string comparison and not the `#` method, store these as class variables so that it's easier to edit them.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
-                        identifier=self.html_ids["Rank"],
-                        title="Change police rank? Must be a number from 0-9",
-                        default={}
-                    )
-                ]
-            )
-        ]
+        default_counter = 0
+        for idx, ele in enumerate(answer):
+            if ele[-8:] == "=default":
```
jsyiek, line 152:
```suggestion
            if ele.endswith("=default"):
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
-                        pseudonym_list_identifier="CorePlugin_assassin_pseudonym",
-                        identifier=self.html_ids["Rank"],
-                        title="Change police rank? Must be a number from 0-9",
-                        default={}
-                    )
-                ]
-            )
-        ]
+        default_counter = 0
+        for idx, ele in enumerate(answer):
```
jsyiek, line 151:
```suggestion
        for idx, ele in enumerate(answer):
            # remove user-added whitespace on either side of their submission
            ele = ele.strip()
```

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
+            if ele[-8:] == "=default":
+                default_counter += 1
+                default = idx
+                answer[idx] = ele[:-8]
+        if default_counter == 0:
```
jsyiek, line 156: In this case, I think it would be sensible to choose the lowest rank and use that as the default.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
+        self.gsdb_set("Manual Rank", not self.gsdb_get('Manual Rank'))
+        return [Label(f"[POLICE] Toggled Manual Ranking from {not self.gsdb_get('Manual Rank')} to {self.gsdb_get('Manual Rank')}")]
+
+    def answer_toggle_police_kills_rankup(self, _):
+        self.gsdb_set("Police Kills Rankup", not self.gsdb_get('Police Kills Rankup'))
+        return [Label(f"[POLICE] Toggled Auto Ranking on Police Kills from {not self.gsdb_get('Police Kills Rankup')} to {self.gsdb_get('Police Kills Rankup')}")]
+
+    def on_event_request_create(self) -> List[HTMLComponent]:
+        # TODO Make a selector with police filtering
+        if self.gsdb_get("Manual Rank"):
```
jsyiek, line 181: nit: you don't need an if/else when both paths are `return`s - you can drop the `else` as it's redundant. In this case, I'd negate the `if` condition, move `return []` into that branch, and dedent the dependency return.

Applies elsewhere as well.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -88,122 +93,205 @@ def __init__(self):
...
             death_manager.add_event(e)
 
+        # Code to generate new ranks if police are promoted/demoted more than ever before
+        if police_rank_manager.get_min_rank() < -int(self.gsdb_get("Default Rank")):
+            current_ranks = self.gsdb_get("Ranks")
+            current_default = int(self.gsdb_get("Default Rank"))
+            for i in range(-current_default - police_rank_manager.get_min_rank()):
+                current_ranks.insert(0, f"Level {-(i+current_default+1)} Constable")
+            self.gsdb_set("Ranks", current_ranks)
+            self.gsdb_set("Default Rank", -police_rank_manager.get_min_rank())
```
jsyiek, line 250: I don't think you need to commit more ranks to the database. How about using a function like this to read from `ranks` to avoid having to mess with the DB?

```python
def getRank(idx: int, ranks: List[str]) -> str:
    if idx >= len(ranks) - 2:
        return ranks[len(ranks) - 3]
    elif idx < 0:
        return ranks[0]
    return ranks[idx]
```

jsyiek (commented):
Mostly minor comments. Well done :) Respond to / address the comments here and I'll approve.

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: The idea behind adding them to the DB was that it would be obvious which new ranks need to be renamed when the umpire configures ranks again, while still ranking up police automatically if the umpire forgets to create the ranks in advance

jsyiek: I see, that makes sense... could you please give a warning to the umpire (as a LABEL etc.) about this and leave a comment in code?

AU2/plugins/custom_plugins/PolicePlugin.py
NerdyNinja11: I've implemented this, but I'm not sure how worth it it is trying to prevent umpires breaking the system by manually editing the json poorly. There are so many instances of things that are expected to be ints or floats or datetimes that would break if a rogue umpire changes them in json to, say, strings; it would be a lot of effort to catch and fix every one of these potential issues, for very little gain

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -95,142 +97,134 @@ def __init__(self):
...
+                                                                             self.plugin_state[plugin_state_id]['default'])
 
     def gsdb_set(self, plugin_state_id, data):
         GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {})[self.plugin_state[plugin_state_id]['id']] = data
 
     def ask_set_ranks(self):
-        # TODO Make Ctrl-C work - not sure why it isn't --> Check Jamie's pull
+        question = []
         default_text = self.gsdb_get("Ranks").copy()
+        if int(self.gsdb_get("Default Rank")) not in range(len(default_text)-2):
```
jsyiek, line 119:
```suggestion
        if not (0 <= int(self.gsdb_get("Default Rank")) < len(default_text) - 2):
```

jsyiek (approved):
Consider changing that if-statement (difference of two integer comparisons vs. scanning a list), but otw LGTM.

NerdyNinja11:
I'm pretty sure that python is very quick testing if something is in a range() object (it's much more optimised than a general list), so I don't think it matters performance-wise. I did it because it was shorter

PR #27, opened 2024-09-28
Add `noInteraction` attribute to `HTMLComponent`, skip `noInteraction` elements in `inquirer_cli`

jsyiek:
Any use of `config` settings will [insert a `HiddenTextbox`](https://github.com/jsyiek/AU2/blob/master/AU2/plugins/CorePlugin.py#L434) at the start of the HTMLComponents that identifies the plugin. This is because config is handled through an export in the `CorePlugin` and not natively in the front-end.

These are passed over by the Inquirer front-end, which means we need to force it to skip it when going backward. The code wasn't doing this.

Resolves bug found in https://github.com/jsyiek/AU2/pull/26


NerdyNinja11 (approved):
Works for me

PR #33, opened 2024-10-19
Attempt/Assist tracking and automatic competency

NerdyNinja11:
Seems to work well, and other than secret word competency and incobash competency, should mean you only have to directly modify competency when modifying events.


AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -69,13 +71,16 @@ def __init__(self):
...
+            "Datetime": self.identifier + "_datetime",
+            "Auto Competency": self.identifier + "_auto_competency"
         }
 
         self.plugin_state = {
             "GAME START": ID_GAME_START,
             "DEFAULT": ID_DEFAULT_EXTN,
-            "COMPETENCY": "competency"
+            "COMPETENCY": "competency",
+            "Auto Competency": "auto_competency",
```
jsyiek, line 82:
```suggestion
            "AUTO COMPETENCY": "auto_competency",

```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -109,6 +120,26 @@ def set_default_competency_deadline_answer(self, htmlResponse):
...
+                options=self.AUTO_COMPETENCY_OPTIONS,
+                selected=GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["Auto Competency"], "Manual")
+            )
+        ]
+
+    def answer_auto_competency(self, htmlResponse):
+        mode = htmlResponse[self.html_ids["Auto Competency"]]
+        GENERIC_STATE_DATABASE.arb_state[self.plugin_state["Auto Competency"]] = mode
+        response = [Label(f"[COMPETENCY] Auto Competency Mode set to {mode}")]
+        if mode != "Manual" and not GENERIC_STATE_DATABASE.plugin_map.get("AttemptPlugin", False):
```
jsyiek, line 137: The config parameter for the `AttemptPlugin` is in the `CompetencyPlugin`. This makes me think we do not need the `AttemptPlugin` and that this functionality can be in the `CompetencyPlugin`. When the parameter is set to an appropriate mode, the `CompetencyPlugin` can avoid asking the `Attempts` question?

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -139,26 +170,71 @@ def on_hook_respond(self, hook: str, htmlResponse, data) -> List[HTMLComponent]:
...
-                        pseudonym_list_identifier="CorePlugin_assassin_pseudonym",
-                        identifier=self.html_ids["Competency"],
-                        title="Extend competency?",
-                        default={},
-                        global_default=GENERIC_STATE_DATABASE.arb_int_state.get(self.plugin_state["DEFAULT"], DEFAULT_EXTENSION)
-                    )
-                ]
-            )
-        ]
+        if GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["Auto Competency"], "Manual") != "Full Auto":
```
jsyiek, line 173: Statements of the form

```
if condition:
    return A
else:
   return B
```

can be simplified to

```
if condition:
    return A
return B
```

In general, you want B to be the longer statement with more nesting.

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -139,26 +170,71 @@ def on_hook_respond(self, hook: str, htmlResponse, data) -> List[HTMLComponent]:
...
+                if not ASSASSINS_DATABASE.get(victim).is_police:
+                    if killer not in e.pluginState[self.identifier][self.plugin_state["COMPETENCY"]]:
+                        # if to avoid overwriting manual entries
+                        message.append(Label(f"[COMPETENCY] Granted {competency_extension} days competency to {killer} for a kill"))
+                        e.pluginState[self.identifier][self.plugin_state["COMPETENCY"]][killer] = competency_extension
+
+            # auto attempt competency:
+            if GENERIC_STATE_DATABASE.plugin_map.get("AttemptPlugin", False):
+                # Is this reliable? It works, but seems to depend on AttemptPlugin.on_event_create() running before this
+                # If not, how would I do this?
```
jsyiek, line 211: If it relies on plugin call ordering, that is determined by a hash-map ordering at initialization time so I wouldn't rely on this.

Two options:
1) The `CompetencyManager` can (and probably should) track this.
2) Merge the `AttemptPlugin` into the `CompetencyPlugin`

I'm partial to (1) - if you delete an earlier event after registering attempts on a later one, the second approach (or the current approach) don't update later events. Not that it matters much, but it's better complexity wise as you aren't looping over all events on adding every event, but tracking `attempts_since_last_kill` dynamically.

This resolves the comment you have later about issues with modifying.

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -139,26 +170,71 @@ def on_hook_respond(self, hook: str, htmlResponse, data) -> List[HTMLComponent]:
...
 
     def on_event_create(self, e: Event, htmlResponse) -> List[HTMLComponent]:
-        e.pluginState.setdefault(self.identifier, {})[self.plugin_state["COMPETENCY"]] = htmlResponse[self.html_ids["Competency"]]
-        return [Label("[COMPETENCY] Success!")]
+        message = []
+        competency_extension = GENERIC_STATE_DATABASE.arb_int_state.get(self.plugin_state["DEFAULT"], DEFAULT_EXTENSION)
+        if GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["Auto Competency"], "Manual") != "Full Auto":
+            e.pluginState.setdefault(self.identifier, {})[self.plugin_state["COMPETENCY"]] = htmlResponse[self.html_ids["Competency"]]
+        else:  # Create an empty competency dict when nothing was entered (due to full auto mode)
+            e.pluginState.setdefault(self.identifier, {})[self.plugin_state["COMPETENCY"]] = {}
```
jsyiek, line 197: AU2 should already be robust to plugins being off. 

```suggestion
```

If you really need it setdefault through all control paths, move the body of the `else` clause before the `if` and update the `if` clause to be `e.pluginState[self.identifier][self. ...] = htmlResponse[...]`

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -139,26 +170,71 @@ def on_hook_respond(self, hook: str, htmlResponse, data) -> List[HTMLComponent]:
...
+                attempts_since_last_kill = {}
+
+                for event in events:
+                    for i in event.pluginState.get("AttemptPlugin", []):
+                        attempts_since_last_kill.setdefault(i, 0)
+                        attempts_since_last_kill[i] += 1
+                    for (killer, victim) in event.kills:
+                        if not ASSASSINS_DATABASE.get(victim).is_police:
+                            attempts_since_last_kill[killer] = 0
+                            # Rules unclear - should attempts be reset upon competency gain (from a kill)?
```
jsyiek, line 223: Doesn't matter so long as you document it. If you're unsure, make it a config parameter.

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -219,6 +219,7 @@ def on_event_request_update(self, e: Event) -> List[HTMLComponent]:
     def on_event_update(self, e: Event, htmlResponse) -> List[HTMLComponent]:
         if not self.gsdb_get("Manual Rank"):
             return [Label("[POLICE] Didn't need to do anything")]
+
```
jsyiek, line 222: nit

```suggestion
```

jsyiek (changes requested):
Good start, but a bit of rework is needed

AU2/plugins/custom_plugins/CompetencyPlugin.py
NerdyNinja11: Is there a reason for this? The former seems more readable to me, and surely the performance cost is negligible

jsyiek: It doesn't end up being more readable because you put heavily indented code at another indentation level more than it needs to be. It gets worse when your else clause is very long, e.g.

```
if skip_work:
    return True
else:
   very
   long
   work
   follows
```

If you then have several nested If statements in the else that could be simplified, you're adding unnecessary indentation and making it harder to follow control flow

jsyiek: (there is no performance cost, it's entirely cosmetic - they almost certainly compile to the same thing in Python bytecode)

NerdyNinja11: It seems to be a debated topic on stackoverflow at least. I'll go with your preference though for consistency

AU2/plugins/custom_plugins/CompetencyPlugin.py
NerdyNinja11: Another problem with (1) I realised as I'm implementing it:
Player gets kill (or 2 attempts), giving 7 days competency.
6 days later, umpires reduce comp time to 5 days.
Player is now inco, if competencymanager tries to dynamically calculate competency.
Should I revert to (2), which comes with the inability to deal with event edits/deletes?
Is there a way of doing everything without storing default competency changes in an event? Or should I just compromise.

AU2/plugins/custom_plugins/CompetencyPlugin.py
NerdyNinja11: I'll just document it

AU2/plugins/custom_plugins/CompetencyPlugin.py
jsyiek: Good point. 

Store the current competency period alongside each attempt. When dynamically calculating it, you'll have full information.

jsyiek:
> Do I need to check that a player is both alive and not police before automatically giving competency, or should this all be handled anyway downstream?

I think it's safe to assume this will be handled downstream.

AU2/plugins/util/CompetencyManager.py
```diff
@@ -17,19 +17,52 @@ class CompetencyManager:
...
+        self.auto_competency = auto_competency
+        self.attempts_since_kill = defaultdict(lambda: 0)
 
     def add_event(self, e: Event):
         for (aID, extn) in e.pluginState.get("CompetencyPlugin", {}).get("competency", {}).items():
             self.deadlines[aID] = max(
                 self.deadlines[aID],
                 e.datetime + datetime.timedelta(days=extn)
             )
+        if self.auto_competency:
```
jsyiek, line 35: you can de-nest this using tricks discussed in previous PR

AU2/plugins/util/CompetencyManager.py
```diff
@@ -17,19 +17,52 @@ class CompetencyManager:
...
 
-    def __init__(self, game_start: datetime.datetime):
+    def __init__(self, game_start: datetime.datetime, auto_competency: bool = False):
         # from assassin ID to deadline
         self.deadlines = defaultdict(lambda: self.game_start + self.initial_competency_period)
         self.game_start = game_start
         self.initial_competency_period = datetime.timedelta(days=GENERIC_STATE_DATABASE.arb_int_state.get(ID_GAME_START, DEFAULT_START_COMPETENCY))
         self.activated = GENERIC_STATE_DATABASE.plugin_map.get("CompetencyPlugin", False)
+        self.auto_competency = auto_competency
+        self.attempts_since_kill = defaultdict(lambda: 0)
```
jsyiek, line 27:
```suggestion
        self.attempts_since_kill = defaultdict(int)

```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -264,8 +299,9 @@ def on_page_generate(self, htmlResponse) -> List[HTMLComponent]:
         events = list(EVENTS_DATABASE.events.values())
         events.sort(key=lambda event: event.datetime)
         start_datetime: datetime.datetime = get_game_start()
+        auto_competency_bool = False if GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["AUTO COMPETENCY"], "Manual") == "Manual" else True
```
jsyiek, line 302:
```suggestion
        auto_competency_bool = GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["AUTO COMPETENCY"], "Manual") != "Manual"

```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -184,58 +224,36 @@ def on_event_request_create(self) -> List[HTMLComponent]:
...
-                        if i not in e.pluginState[self.identifier][self.plugin_state["COMPETENCY"]]:
-                            message.append(Label(f"[COMPETENCY] Granted {competency_extension} days competency to {i} for attempts"))
-                            e.pluginState[self.identifier][self.plugin_state["COMPETENCY"]][i] = competency_extension
-
-        if not message:
-            message.append(Label("[COMPETENCY] Success!"))
-        return message
+        if self.html_ids["Attempts"] in htmlResponse:
+            e.pluginState.setdefault(self.identifier, {})[self.plugin_state["ATTEMPTS"]] = htmlResponse[self.html_ids["Attempts"]]
+        # Store the default competency extension aat the time of the event, in the event
```
jsyiek, line 248:
```suggestion
        # Store the default competency extension at the time of the event, in the event

```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -95,9 +108,30 @@ def __init__(self):
...
+                answer=self.answer_toggle_attempt_tracking
             )
         ]
 
+    def answer_test_incos(self, _):  # TODO Remove before PR gets merged
+        events = list(EVENTS_DATABASE.events.values())
+        events.sort(key=lambda event: event.datetime)
+        start_datetime: datetime.datetime = get_game_start()
+        auto_competency_bool = False if GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["AUTO COMPETENCY"],
+                                                                             "Manual") == "Manual" else True
```
jsyiek, line 125:
```suggestion
        auto_competency_bool = GENERIC_STATE_DATABASE.arb_state.get(self.plugin_state["AUTO COMPETENCY"], "Manual") != "Manual"

```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -95,9 +108,30 @@ def __init__(self):
...
+            ),
+            ConfigExport(
+                identifier="CompetencyPlugin_attempt_tracking",
+                display_name="Competency -> Toggle Attempt Tracking",
+                ask=lambda *args: [],
+                answer=self.answer_toggle_attempt_tracking
             )
         ]
 
+    def answer_test_incos(self, _):  # TODO Remove before PR gets merged
```
jsyiek, line 120:
```
# TODO Remove before PR gets merged
```

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -72,17 +73,29 @@ def __init__(self):
...
+            "AUTO COMPETENCY": "auto_competency",
+            "ATTEMPT TRACKING": "attempt_tracking",
             "COMPETENCY": "competency",
-            "Auto Competency": "auto_competency",
-
+            "ATTEMPTS": "attempts",
+            "CURRENT DEFAULT": "current_default"
         }
 
+        self.exports = [  # TODO Remove before PR gets merged and I don't have to test everything anymore
```
jsyiek, line 90:
```
 # TODO Remove before PR gets merged
```

jsyiek (approved):
resolve conflicts, address comments, and you can merge

PR #47, opened 2024-10-29
[BUG] Double deaths cause `TargetingPlugin.compute_targets` to crash

jsyiek:
image

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -350,6 +350,15 @@ class TargetingPlugin(AbstractPlugin):
         The problem is constraint satisfaction so a Prolog-style "generate-and-test" is sufficient.
         """
 
+        deaths_adj = set([d for d in deaths if d in targeting_graph])
+        visited = []
+        for d in deaths:
+            if d not in deaths_adj or d in visited:
+                print(f"WARNING: {d} has been killed more than once. Skipping this player when updating graph...")
+            visited.append(d)
+
+        deaths = deaths_adj
+
         # collect a list of [non-unique] players who need new targets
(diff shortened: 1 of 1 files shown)


AU2/plugins/custom_plugins/TargetingPlugin.py
```diff
@@ -350,6 +350,15 @@ def update_graph(
         The problem is constraint satisfaction so a Prolog-style "generate-and-test" is sufficient.
         """
 
+        deaths_adj = list(set([d for d in deaths if d in targeting_graph]))
+        visited = []
+        for d in deaths:
+            if d not in deaths_adj or d in visited:
+                print(f"WARNING: {d} has been killed more than once. Skipping this player when updating graph...")
```
NerdyNinja11, line 357: Do we do print statements? I've been returning Labels for warnings to the umpires (compatibilty for a future html frontend and all that)
Unless you do want this to just be a console warning

jsyiek:
I'm going to add a regression test tomorrow to this PR - but curious for thoughts on the fix

NerdyNinja11 (approved):
Looks good. Probably just double check with the umpires that this does, in fact, generate the same graph before any emails are sent

AU2/plugins/custom_plugins/TargetingPlugin.py
jsyiek: Valid point.

jsyiek: I've got an idea for fixing this. Let's get the fix to prod, and then I'll spin up another PR with some new library functions.

jsyiek:
ah, crap, I committed the tests directly to master. no matter. they'd have gone directly there anyway.

PR #48, opened 2024-10-30
Print using labels in TargetingPlugin

jsyiek:
**What is it**: A simple library class that can be used to emulate `print`-like behavior, but converts into a list of `Label`s behind the scene which can be accessed with `instance.get_prints()`

**What's changed**: Only `TargetingPlugin`, which was using raw Python `print`'s. As @NerdyNinja11 mentioned, this is not scalable with web UIs. Note that while the `SRCFPlugin` uses prints, I plan to disable this function in the web UI as the web server should run on the SRCF (so doesn't need an SSH/SFTP client). 

**Testing**: I've printed out the targeting graph and tested its associated functions.

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -80,9 +80,11 @@ class TargetingPlugin(AbstractPlugin):
         if hook == "SRCFPlugin_email":
             # if graph computation time becomes an issue, we could yield the `last_graph` and then compute
             # current_graph without needing to recompute
-            last_emailed_event = int(GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {}).setdefault("last_emailed_event", -1))
-            last_graph = self.compute_targets(max_event=last_emailed_event)
-            current_graph = self.compute_targets()
+            response = []
+            last_emailed_event = int(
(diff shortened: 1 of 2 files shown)


NerdyNinja11 (commented):
Seems a whole new class is a little overengineered for just the TargetingPlugin.
It feels like abstraction for the sake of abstraction to me, since imo just having a list and appending `Label`s to that directly is easier to understand.

Additionally, are you ever returning the list of labels for `update_graph` or `compute_targets` to actually be displayed?

jsyiek:
gahhh I missed that... I tested it and then edited the API slightly (originally had it as an instance variable rather than a local variable) and forgot to adjust the return - great spot. 

I agree slightly, but I feel the intent is clearer than an ambiguous list of Labels (and would look a bit less cluttered in the code page). would also allow for more advanced formatting later if we need it. Do these arguments sway you at all?

NerdyNinja11:
Makes sense. I personally am not a fan of stuff that uses extra classes/imports (when it could be done just as easily without), as when trying to work out what the code does for the first time, I have to open and read a whole new file rather than just knowing how a list behaves already (with maybe an explanatory comment.
On the other hand, this is a niche opinion that pretty much every code person I've talked to disagrees with, so I suppose it's probably fine, and, as you say, easier to expand later.
It would probably be worth then using this in every instance where I've conditionally returned Labels

NerdyNinja11 (approved):

Issue #4, opened 2024-06-13
Pseudonyms not rendering correctly on death if later pseudonym added


jsyiek:
Steps to reproduce:
- Add an assassin
- Add an event with their death
- Add another pseudonym to that assassin

thenewtonium:
So, to clarify, the bug here is that [DX] renders the later pseudonym as well?

jsyiek:
Yup. The cause comes down in part to AU2's design philosophy. AU1 would update the state throughout the game, meaning bugs that cause the state to be incorrectly stored or make the state inconsistent would halt the game (e.g., the targeting graph is stored in full, preventing the annulment of a kill). AU2 attempted to rectify this by trying to avoid storing this derived state and only storing the starting points from where it could be regenerated later. 

To this end, events don't store the pseudonyms you had at the time (in case you change them later or when the event occurred). A simple solution at first seems like it could be just to store the pseudonyms the assassins had (or how many, etc.) but it introduces issues if the umpire goes to update the event/assassin later and could lead to inconsistency (e.g., storing a pseudonym becomes inconsistent if the umpire later edits it).

I'm unsure how to fix this other than making a changelog to each assassin's pseudonym list in `Assassin` and using event datetime to determine which pseudonyms existed at death time. I'm not necessarily a big fan of this solution. For one, it would need to come with an appropriate setter method for `pseudonyms` so that future plugins that want to add/edit/delete/etc. pseudonyms on their own and not through `CorePlugin` correctly update the change log. 

It's not simple and only occurs in May Week and in niche cases with police officers who die far too many times (rare).

thenewtonium:
What is the issue with storing the number of pseudonyms a player had at the time? It's no less consistent than the way the main pseudonym is stored, surely?

jsyiek:
Yeah, part of the reason this issue has sat open for a long time is every solution I've thought up has some weird edge case that violates state consistency requirements that could plausibly be committed by an umpire within the next 15 years.

PR #49, opened 2024-11-01
[4] Store datetime that each pseudonym is added, in order to render [DX] non-anachronistically

thenewtonium:
Firstly, to store the datetime of each pseudonym I have added a `dict` field `pseudonym_datetimes` to the `Assassin` class, which maps the index of a pseudonym to the datetime it was added, and have used this to filter out pseudonyms added after an event happened when rendering [DX] in headlines/reports.
The reason for doing it this way is to be backwards-compatible with old databases files, as it makes the datetimes optional. Any pseudonym without a corresponding datetime in this `dict` is assumed to have existed "forever", so will always appear in the rendering of [DX].

Next, using indices to refer to pseudonyms in this way obviously means that we can't actually delete pseudonyms because it may change which pseudonym a given index refers to. (This was just generally an issue anyway, since the mapping from assassins to pseudonyms stored in events can also be messed up by deleting pseudonyms...). Hence, I have made it so that you cannot actually delete a pseudonym, only hide it — this is done by setting it to be blank. Moreover, the initial pseudonym cannot be deleted.

However, the initial pseudonym (and all the others) can be edited: the final change I have made (chronologically actually the first, but it's the conclusion of the flow of logic I am presenting here) is to the interface for editing a player's pseudonyms. Instead of deselecting any pseudonyms to delete and selecting `*NEW*` to trigger a text input for a list of new pseudonyms, the new interface (for which I created a new `EditablePseudonymList` HTMLcomponent) instead prompts the user to select a pseudonym to edit, or to select `*NEW*` to add a new pseudonym. After editing or adding a new pseudonym the user is returned to this menu; there is additionally a `*CONTINUE*` option to move on from this menu.
Moreover, rather than simply returning the final set of pseudonyms to `CorePlugin`, this interface returns, separately

- A `list` of new pseudonyms
- A `dict` mapping pseudonyms that have been edited (referenced by index) to their new values
- A `set` of pseudonyms to delete (referenced by index)

`CorePlugin` then processes these outputs using relevant getter and setter methods in `Assassin`.


thenewtonium:
(This is still a work in progress — need to find a way of tying the datetime of a pseudonym to an event rather than using when the umpire adds it to the system!)

jsyiek:
I think it's valid to treat a pseudonym as canon in the timeline at the point the umpire adds it - or allow the umpire optionally to specify from when it's valid.

thenewtonium:
The issue I was thinking of is that is certainly possible for a player to make a second kill and claim a new pseudonym, make an attempt/kill under that pseudonym, then get killed, all before the umpire has managed to update the system.

But allowing an umpire to specify validity does solve this issue...
To make the inteface nicer it might make sense to move pseudonym management out of `Assassin -> Update` and into a seperate plugin export if the pseudonym editing interface is going to have to deal with both the values *and* validities of pseudonyms. This would also have the benefit that the umpire doesn't have to mash <kbd>Enter</kbd> after adding a new pseudonym to get through all the other assassin data that they don't want to change. What do you think?

jsyiek:
I was actually thinking this as well tbh. I've wondered about whether there is design space for a `GameUtils` plugin which would add a bunch of convenience tools not strictly needed for game success. One such export could be `Assassin -> Quick Update`, which uses the secret third step `options` of the `ask -> answer` to edit only one field in the ask.

AU2/frontends/inquirer_cli.py
```diff
@@ -429,6 +445,104 @@ def render(html_component, dependency_context={}):
             inquirer.Text(name=html_component.identifier, message=html_component.title, default=html_component.default)]
         return inquirer_prompt_with_abort(q)
 
+    elif isinstance(html_component, EditablePseudonymList):
+        # perhaps abstract this?
+        # so that references to pseudonyms and datetimes are contained within attributes of html_component,
+        # rather than being hard-coded here.
```
jsyiek, line 451: Don't follow - can you elaborate?

AU2/html_components/EditablePseudonymList.py
```diff
@@ -0,0 +1,25 @@
...
+        self.uniqueStr = self.get_unique_str()
+        # values should be a list of tuples giving pseudonym text + validity start datetime
+        self.values = [a for a in values]
+        super().__init__()
+
+    def _representation(self) -> str:
+        # TODO: create a HTML version of this component.
+        #       Broadly, it should be a list of textboxes with the list values given,
+        #       with the option to add new textboxes.
+        raise NotImplementedError()
```
jsyiek, line 25: missing newline at end of file

AU2/frontends/inquirer_cli.py
```diff
@@ -429,6 +445,104 @@ def render(html_component, dependency_context={}):
...
+                        edited[c] = p_data
+                    else:
+                        new_values[c-old_n_values] = p_data
+
+        return {html_component.identifier:
+                    ListUpdates(edited, new_values, deleted_indices)}
+
     elif isinstance(html_component, ArbitraryList):
         q = [inquirer.Checkbox(
             name=html_component.identifier,
```
jsyiek, line 548: Please can you delete `ArbitraryList` and all references to it given your refactor into `EditablePseudonymList`? `ArbitraryList` is very badly named but appears to only apply to pseudonyms (it looks like it was meant to be for a list for which things can be added arbitrarily, but wasn't generalized due to time constraints in MW24 development)

jsyiek (changes requested):
good overall, some minor comments

jsyiek:
Have you tested that pseudonyms render correctly on death (i.e., that the datetimes are used correctly)? Just to confirm - I presume you have

AU2/database/model/Assassin.py
```diff
@@ -42,9 +58,55 @@ def __post_init__(self):
             self.identifier = f"{self.real_name} ({self.pseudonyms[0][:15]}{dotdotdot}) ID: {self._secret_id}"
 
     def get_pseudonym(self, i: int) -> str:
-        if i >= len(self.pseudonyms):
-            return self.pseudonyms[-1]
+        if i >= len(self.pseudonyms) or not self.pseudonyms[i]:
+            # fallback to last non-null pseudonym
```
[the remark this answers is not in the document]
thenewtonium: Yep; although actually deletion sets them to `""` rather than `None`, because the latter gets JSONified as the *string* `"None"` rather than an actual null value... (So I should probably clarify this in the comment as well!)

AU2/frontends/inquirer_cli.py
thenewtonium: My thinking here was that fundamentally all that is going on in this interface is editing a list where each entry has multiple parts to its value, so the component doesn't need to be specfically tailored to being the pseudonym editor.
Although, if we do want to reuse the component in another context it probably makes more sense to cross that bridge when we come to it rather than pre-emptively abstracting...

thenewtonium:
> Have you tested that pseudonyms render correctly on death (i.e., that the datetimes are used correctly)? Just to confirm - I presume you have

I had tested that pseudonyms render correctly in the sense of solving #4, but this reminded me that I hadn't tested what happens to the author of a report when a pseudonym is deleted. It turned out I'd missed changing an `assassins_model.pseudonyms[pseudonym_index]` to an `assassins_model.get_pseudonym(pseudonym_index)` so it did not render correctly in that case. (I've fixed that in [`da3f717`](https://github.com/jsyiek/AU2/pull/49/commits/da3f7a7db705e169e6d7210c3809924a3cff7177) now though.)

AU2/frontends/inquirer_cli.py
jsyiek: I see, thank you. That sounds like a sensible note for future. Perhaps leave a `# TODO: ...` with some of what you have here as it's a good description of the potential improvement.

jsyiek:
As this PR has lots of commits, please can you squash them?

```
git rebase -i master # then change everything from `pick` to `s` (and choose a short snappy commit msg)
git push -f
```

AU2/database/model/Assassin.py
```diff
@@ -42,9 +58,104 @@ def __post_init__(self):
...
+    def delete_pseudonym(self, i: int):
+        """
+        Deletes a pseudonym, referenced by index, from this assassin.
+        Internally, the pseudonym is set to "" so that the indices of other pseudonyms are unchanged.
+        The initial pseudonym (i=0) cannot be deleted.
+
+        :param i: the index of the pseudonym to delete
+        """
+        if i == 0:
+            raise ValueError("Cannot delete initial pseudonym.")
```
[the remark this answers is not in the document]
thenewtonium: Hmm... Yes I think it is probably better for the program not to crash over a silly error. My one concern is that it may make debugging more time-consuming as the source of weird behaviour will be less clear than if the program just raises an exception. I'm comfortable with it possibly failing silently in this case though since the "weird" behaviour here is just a guardrail against the umpire doing something silly.

Just to clarify: this `Try` doesn't already exist and would need to be declared somewhere in AU2? I can't find any reference to it anywhere.

Issue #19, opened 2024-06-16
Escape `{` and `}` in reports


jsyiek:
![image](https://github.com/jsyiek/AU2/assets/43001551/e9b98561-4681-47a8-bfc0-5a1ed6ab6732)

jsyiek:
format strings don't play well with `{` in reports

jsyiek:
could be fixed by `.replace`'ing the `{` or `}` with their html code equivalents? the issue is that all pseudonyms/heads/news are escaped (which would kill the escape), so it's not a complete solution

NerdyNinja11:
'If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}' - python docs.
Would this work as initial sanitisation to reports/headlines?

jsyiek:
> 'If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}' - python docs.
> 
> Would this work as initial sanitisation to reports/headlines?

Probably! All text inputs should sanitize in the front-end with a .replace call.

thenewtonium:
> 'If you need to include a brace character in the literal text, it can be escaped by doubling: {{ and }}' - python docs. Would this work as initial sanitisation to reports/headlines?

Does this not cause issues with page generation though? Since there the relevant strings do not get passed through `.format` and so will be rendered with double rather than single braces.

Since it seems to only be `inquirer` the thas an issue with curly braces, might it not make more sense to escape strings being passed into this instead? PR #51 demonstrates what I mean.

jsyiek:
Yeah, this is what I was saying about sanitizing in the front-end only.

PR #51, opened 2024-11-06
[19] Escaped {} in inquirer arguments

thenewtonium:
This PR implements escaping of `message` and `default` arguments of `inquirer` Questions to solve issue #19.
I have tested the entire assassin create / update and event create / update interfaces with values including {}, and there were no issues. Pages also generated correctly.

### DIFF
--- a/AU2/plugins/util/game.py
+++ b/AU2/plugins/util/game.py
@@ -43,6 +43,16 @@ def escape_format_braces(string: str) -> str:
     Escapes { and } in a string so that they will be processed correctly by .format
     This needs to be called on user-input strings passed as part of the message of any inquirer prompt,
     or as part of the default of inquirer Text prompts (but not the choices or defaults of List of Checkbox prompts)
+
+    Args:
+        string: the string to escape occurrences of `{` and `}` in.
+
+    Returns:
+        `string` with all occurrences of `{` and `}` doubled.
+
+    Examples:
+        >>> escape_format_braces(":} :{")
+        ":}} :{{"
     """
     # need type check because, when escaping default values, `None` is sometimes passed to this function
-    return string.replace("{", "{{").replace("}", "}}") if isinstance(string, str) else string
+    return string.replace("{", "{{").replace("}", "}}")
--- a/AU2/plugins/util/game.py
+++ b/AU2/plugins/util/game.py
@@ -36,3 +36,13 @@ def soft_escape(string: str) -> str:
     if not string.startswith("<!--HTML-->"):
         return escape(string)
     return string
+
+
+def escape_format_braces(string: str) -> str:
+    """
+    Escapes { and } in a string so that they will be processed correctly by .format
+    This needs to be called on user-input strings passed as part of the message of any inquirer prompt,
+    or as part of the default of inquirer Text prompts (but not the choices or defaults of List of Checkbox prompts)
+    """
+    # need type check because, when escaping default values, `None` is sometimes passed to this function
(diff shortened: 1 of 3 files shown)


AU2/plugins/util/game.py
```diff
@@ -35,4 +35,14 @@ def soft_escape(string: str) -> str:
...
+
+
+def escape_format_braces(string: str) -> str:
+    """
+    Escapes { and } in a string so that they will be processed correctly by .format
+    This needs to be called on user-input strings passed as part of the message of any inquirer prompt,
+    or as part of the default of inquirer Text prompts (but not the choices or defaults of List of Checkbox prompts)
+    """
+    # need type check because, when escaping default values, `None` is sometimes passed to this function
+    return string.replace("{", "{{").replace("}", "}}") if isinstance(string, str) else string
```
jsyiek, line 48: why would this not be a string? require inputs to be cast to a `str` or cast them to a str in the function (and take `Any` as the input parameter type)

AU2/plugins/util/game.py
```diff
@@ -35,4 +35,14 @@ def soft_escape(string: str) -> str:
...
\ No newline at end of file
+    return string
+
+
+def escape_format_braces(string: str) -> str:
+    """
+    Escapes { and } in a string so that they will be processed correctly by .format
+    This needs to be called on user-input strings passed as part of the message of any inquirer prompt,
+    or as part of the default of inquirer Text prompts (but not the choices or defaults of List of Checkbox prompts)
+    """
```
jsyiek, line 46: Can you finish the doc-string with params & return? I need to clean up the rest of this file - most of it dates back to the days before MW24 where speed was needed over style.

My preferred style is [Google's doc string style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), e.g.:

```
def function_with_types_in_docstring(param1, param2):
    """Example function with types documented in the docstring.

    `PEP 484`_ type annotations are supported. If attribute, parameter, and
    return types are annotated according to `PEP 484`_, they do not need to be
    included in the docstring:

    Args:
        param1 (int): The first parameter.
        param2 (str): The second parameter.

    Returns:
        bool: The return value. True for success, False otherwise.

    .. _PEP 484:
        https://www.python.org/dev/peps/pep-0484/

    """


def function_with_pep484_type_annotations(param1: int, param2: str) -> bool:
    """Example function with PEP 484 type annotations.

    Args:
        param1: The first parameter.
        param2: The second parameter.

    Returns:
        The return value. True for success, False otherwise.

    """
```

jsyiek (commented):
Just one question for you. This PR has brought to my attention the amount of tech debt the MW24 AU2 rush built up... I'll add that to my to-do list to fix.

AU2/plugins/util/game.py
thenewtonium: Sometimes `None` is set as the default value of a `LargeTextEntry` component and so without this check trying to sanitise the value throws an error. Casting to `str` doesn't work for that because it gives the string `'None'`.

I could check explicitly for `None` and cast everything else to `str`. Though, it might be better to just change the default value for `LargeTextEntry` to be "" to not have to deal with non-string values at all, which would also mean we wouldn't have to deal with the `None` case when rendering the HTML version of the component.

thenewtonium: Update: commit [`f0d3afe`](https://github.com/jsyiek/AU2/pull/51/commits/f0d3afe824c0519b1ef55f37f11ef1849b0bbdfe) implements what I talked about above.

jsyiek (approved):
lgtm

Issue #57, opened 2024-11-09
Find event by date


thenewtonium:
Following on from issue #55: the interface for locating an event in `Event -> Update` (and also `Event -> Delete`) gets kind of unwieldy with a lot of events. 

It would be a lot easier to search for an event using its datetime stamp.

At a minimum, the list of events should include the datetime of each event and be sorted by datetime. This is doable within the current plugin framework since `inquirer.List` can have options return different values to their display values, so we just need to change the `options_functions` of the event update export accordingly.

More useful would be a more sophisticated method of selecting an event, for example grouping into sub-lists by date or using something like `Searchable` from PR #56. This would require reworking how plugin exports are handled however.

jsyiek:
Yeah, this would require an entirely different approach to #56 - what if each `option` could take in the results of all previous options?

thenewtonium:
Yeah that does seem like a good approach for grouping by date!

thenewtonium:
I'm not sure if we actually want to implement this anymore, #58 seems to be sufficient and an extra layer (grouping by date or having a search function) might just be more annoying. If we do implement this it should probably somehow be made togglable by `UIConfigPlugin`.

PR #58, opened 2024-11-10
 [57] Display datetimes in event selector of Event -> Update

thenewtonium:
This is the "minimum change" for issue #57.

This PR doesn't change the plugin framework, but changes the `options_functions` of Event -> Update to instead return a list of tuples, where the first part of each tuple is what is displayed to the user and the second part is what will be passed to the plugin's `ask` function. The options are also sorted by event datetime, in reverse chronological order.

The display value of each event is the datetime + truncated headline. This means it no longer includes the internal identifier for the event; I'd be interested in hearing thoughts about this.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -1,7 +1,7 @@
 import glob
 import os.path
 
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
 from AU2.database.EventsDatabase import EVENTS_DATABASE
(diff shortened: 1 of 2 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -145,7 +145,7 @@ def __init__(self):
                 "Event -> Update",
                 self.ask_core_plugin_update_event,
                 self.answer_core_plugin_update_event,
-                (lambda: [v for v in EVENTS_DATABASE.events],)
+                (self.gather_events,)
```
jsyiek, line 148: `Event -> Delete` needs to be updated with this as well

jsyiek:
I don't really think it's an issue for event ID not to be shown... not yet, anyway

jsyiek (approved):
lgtm, I expect the umpires will like this one!

thenewtonium:
(I'm going to keep the branch and issue #57 open because this only implements the first part of the enhancement.)

Issue #36, opened 2024-10-27
Debug utils plugin / API


jsyiek:
There's a big need for something that helps both umpires and developers debug AU2. Such a solution should include at least these ideas I've thought up:
- viewing past email files 
- viewing an assassin's/events summary within AU without having to use the clunky X -> Update 
- viewing the SRCF log files (see SRCF plugin)
- viewing targets for specific players 
Etc

The main challenge with implementing this is that the desirable data to view is tied to specific plugins. Perhaps a framework improvement is needed - maybe CorePlugin has a Debug option and the concept of a "DebugExport" is added?

PR #59, opened 2024-11-11
[44] Added Assassin -> Hide export

thenewtonium:
This allows the umpire to toggle the visibility for each assassin.
An assassin is hidden by setting its `hidden` attribute to `True`.
Hidden assassins are not included in the options for `Assassin -> Update` or in the `AssassinPseudonymPair` component of `Event -> Create`.
Hidden assassins that are *already* in the event are included in the `AssassinPseudonymPair` component of `Event -> Update`, but the rest are not.
Hiding an assassin has no other effects currently. In particular, it has no effect on `TargetingPlugin` or sending emails with `SRCFPlugin`.

### DIFF
--- a/AU2/database/AssassinsDatabase.py
+++ b/AU2/database/AssassinsDatabase.py
@@ -2,7 +2,7 @@ import os
 
 from dataclasses import dataclass
 from dataclasses_json import dataclass_json
-from typing import Dict, List
+from typing import Dict, List, Callable, Tuple, Optional
 
 from AU2 import BASE_WRITE_LOCATION
 from AU2.database.model import PersistentFile, Assassin
@@ -44,12 +44,54 @@ class AssassinsDatabase(PersistentFile):
         """
         return self.assassins[identifier]
 
-    def get_identifiers(self) -> List[str]:
+    def get_filtered(self, *,
+                     include: Callable[[Assassin], bool] = lambda x: True,
+                     include_hidden: Callable[[Assassin], bool] = lambda x: False) -> List[Assassin]:
         """
+        Parameters:
+            include: function that takes a non-hidden assassin and returns whether it should be included in the
+                returned list. Defaults to always return True.
+            include_hidden: function that takes a hidden Assassin object and returns whether it should be include in the
+                returned list. Defaults to always return False.
+
         Returns:
-            list of identifiers sorted alphabetically
+            list of non-hidden assassins (as Assassin objects) satisfying `include`, and hidden assassins satisfying
+                include_hidden
         """
-        return sorted([v for v in self.assassins], key=lambda n: n.lower())
+        return [a for a in self.assassins.values() if (a.hidden and include_hidden(a)) or (not a.hidden and include(a))]
+
+    def get_identifiers(self, *,
+                     include: Callable[[Assassin], bool] = lambda x: True,
+                     include_hidden: Callable[[Assassin], bool] = lambda x: False) -> List[str]:
+        """
+        Parameters:
+            include: function that takes a non-hidden assassin and returns whether it should be included in the
+                returned list. Defaults to always return True.
+            include_hidden: function that takes a hidden Assassin object and returns whether it should be include in the
+                returned list. Defaults to always return False.
+        Returns:
+            list of identifiers sorted alphabetically, filtered according to the `include` and `include_hidden` functions
+        """
+        return sorted([a.identifier for a in self.get_filtered(include=include, include_hidden=include_hidden)])
+
+    def get_ident_pseudonym_pairs(self, *,
+                     include: Callable[[Assassin], bool] = lambda x: True,
+                     include_hidden: Callable[[Assassin], bool] = lambda x: False) -> List[Tuple[str, List[str]]]:
+        """
+        Parameters:
+            include: function that takes a non-hidden assassin and returns whether it should be included in the
+                returned list. Defaults to always return True.
+            include_hidden: function that takes a hidden Assassin object and returns whether it should be include in the
+                returned list. Defaults to always return False.
+        Returns:
+            list of identifiers sorted alphabetically, filtered according to the `include` and `include_hidden` functions
+        """
+        return sorted([
+            (a.identifier, a.pseudonyms) for a in self.get_filtered(
+                include=include,
+                include_hidden=include_hidden
+            )
+        ])
 
     def _refresh(self):
         """
--- a/AU2/database/model/Assassin.py
+++ b/AU2/database/model/Assassin.py
@@ -24,6 +24,10 @@ class Assassin(PersistentFile):
     college: str
     notes: str
     is_police: bool
+    # we should not delete assassins because that will break any references to that assassin.
+    # most significantly, it would mess up the targeting graph.
+    # but it is useful to hide certain assassins in interfaces (e.g. an assassin who has been cloned into police)
+    hidden: bool = False
     # almost everything that is stateful is probably best placed in an event
     # but for a few plugins, it might make sense to place the information directly onto the assassin
     # make sure you know what you're doing (modifications here can't be undone with a later event state change)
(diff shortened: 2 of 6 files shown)


AU2/database/AssassinsDatabase.py
```diff
@@ -44,12 +44,12 @@ def get(self, identifier: str) -> Assassin:
...
         return self.assassins[identifier]
 
-    def get_identifiers(self) -> List[str]:
+    def get_identifiers(self, include_hidden: bool = False) -> List[str]:
         """
         Returns:
             list of identifiers sorted alphabetically
         """
-        return sorted([v for v in self.assassins], key=lambda n: n.lower())
+        return sorted([id for id, a in self.assassins.items() if include_hidden or (not a.hidden)], key=lambda n: n.lower())
```
jsyiek, line 52: you don't need brackets around `not a.hidden`. also, `id` shadows the built-in local function of the same name - could you change this to `identifier` or similar?

AU2/database/model/Assassin.py
```diff
@@ -24,6 +24,9 @@ class Assassin(PersistentFile):
     college: str
     notes: str
     is_police: bool
+    # we cannot delete assassins because that will break thingsm,
```
jsyiek, line 27:
```suggestion
    # we cannot delete assassins because that will break things,
```

Some examples of things maybe? (e.g., events which take in assassin identifiers)

AU2/frontends/inquirer_cli.py
```diff
@@ -166,6 +166,9 @@ def render(html_component, dependency_context={}):
     # dependent component
     elif isinstance(html_component, AssassinPseudonymPair):
         assassins = [a[0] for a in html_component.assassins]
+        # enforce including defaults, in case they are hidden, so that the user doesn't accidentally remove an assassin
```
jsyiek, line 169:
```suggestion
        # context: Searchable meta-component
        # enforce including defaults, in case they are hidden, so that the user doesn't accidentally remove an assassin
```

?

AU2/plugins/CorePlugin.py
```diff
@@ -237,7 +243,7 @@ def on_assassin_update(self, assassin: Assassin, htmlResponse: Dict) -> List[HTM
         return [Label("[CORE] Success!")]
 
     def on_event_request_create(self):
-        assassins = [(a.identifier, a.pseudonyms) for a in ASSASSINS_DATABASE.assassins.values()]
+        assassins = [(a.identifier, a.pseudonyms) for a in ASSASSINS_DATABASE.assassins.values() if not a.hidden]
```
jsyiek, line 246: This should probably be extracted to a library method of `ASSASSINS_DATABASE` given it's also used in `on_event_request_update` - it might end up used elsewhere

AU2/plugins/CorePlugin.py
```diff
@@ -381,6 +385,21 @@ def answer_core_plugin_update_assassin(self, html_response_args: Dict):
             return_components += p.on_assassin_update(assassin, html_response_args)
         return return_components
 
+    def ask_core_plugin_hide_assassins(self):
+        assassins = ASSASSINS_DATABASE.get_identifiers(include_hidden=True)
+        components = [SelectorList(identifier=self.identifier+"_hidden_assassins",
```
jsyiek, line 390: This identifier should probably live in HTML ids

jsyiek (changes requested):
minor comments, and I like the approach used to solve this one (e.g., allowing assassins to be "unhidden" and not calling it delete)

AU2/frontends/inquirer_cli.py
thenewtonium: The context here is actually assassins with `hidden` set to `True`, so I guess I should clarify that (though it also works for `Searchable` as well)

AU2/plugins/CorePlugin.py
thenewtonium: The whole thing? Or just the filtering out of hidden assassins?

AU2/plugins/CorePlugin.py
thenewtonium: I was thinking about doing this originally but I was concerned by the fact that `html_ids` only includes the identifiers for Assassin-relevant components, and the other identifiers are segregated into seperate dicts (`event_html_ids`, `config_html_ids`).

Relatedly, I've been wondering what the purpose of `html_ids_list` is? It doesn't seem to actually be used anywhere and currently just duplicates `html_ids.keys()`.

AU2/frontends/inquirer_cli.py
thenewtonium: Actually, I've realised the component doesn't handle the added defaults properly oops. I thought I'd tested this but I must have <kbd>Ctrl</kbd>+<kbd>C</kbd>'d out every time... To fix this the component would need to query `ASSASSINS_DATABASE` to fetch the missing pseuodnym which doesn't seem like something we want it to do. So I'm going to move the inclusion of defaults into `on_event_request_update` instead

AU2/plugins/CorePlugin.py
jsyiek: There's a good chance this is tech debt built up during the rush to get AU2 finished before MW24. Feel free to delete if there's no need for it.

AU2/plugins/CorePlugin.py
jsyiek: Fetching assassin ID/pseudonym pairs

AU2/database/AssassinsDatabase.py
```diff
@@ -44,12 +44,31 @@ def get(self, identifier: str) -> Assassin:
...
+            include_hidden: function that takes a hidden Assassin object and returns whether it should be include in the
+                returned list. Defaults to always return False.
+
+        Returns:
+            list of Assassin objects satisfying the filter, excluding hidden assassins unless `include_hidden` is set to
+            True
+        """
+        return [a for a in self.assassins.values() if (a.hidden and include_hidden(a)) or (not a.hidden and filter(a))]
+
+    def get_identifiers(self, include_hidden: bool = False) -> List[str]:
```
jsyiek, line 63: This function should be a wrapper around `get_filtered` - i.e., call it with a lambda to avoid code duplication.

AU2/plugins/CorePlugin.py
```diff
@@ -168,8 +170,19 @@ def __init__(self):
...
         ]
 
-        # str -> HookedExport
-        self.hooks = {}
+        self.hooks = {
+            "hide_assassins": self.identifier + "_hide_assassins",
+        }
+
+        self.hooked_exports = [
+            HookedExport(
```
jsyiek, line 178: You can make this an API call exposed directly in `AbstractPlugin` with a default
"Do nothing" implementation, like the similar `on_assassin_create`. `HookedExport`s will work here but they're a little clunkier and mostly intended for plugins that can be enabled or disabled at will (hence not justifying a dedicated API call for their functionality).

jsyiek (changes requested):

AU2/plugins/CorePlugin.py
thenewtonium: The reason I did this as a hook is that, I was thinking that once #63 is implemented there wouldn't be much reason to use this export and we'd want to move it into a debug plugin à la #36. I was also thinking that the only reason it needs a hook at all is for `UIConfigPlugin`, so didn't justify its own call.

But come to think of it, we might want also to a, for example, have `TargetingPlugin` give a warning when hiding live assassins that hidden assassins will still be included in the targeting tree. So maybe it does justify an API call?

jsyiek:
> Hiding an assassin has no other effects currently. In particular, it has no effect on TargetingPlugin or sending emails with SRCFPlugin.

Completely fine not to deal with them now, but would be nice to make the SRCFPlugin ignore hidden assassins when sending emails

jsyiek:
Also, you have merge conflicts.

jsyiek (approved):
LGTM

PR #61, opened 2024-11-11
[60] Bounty plugin

### LINKED ISSUE #60: (URGENT) BountyPlugin
Separate plugin to generate a bounties page (similar to wanted).

jsyiek:
Closes #60 

**Problem**: A player may end up placing a bounty on another player within the next day or so. This one slipped through the cracks of AU2 development as it was previously incorporated into the `MafiaPlugin` (which is now seriously out-of-date) and so didn't get an issue created back during the first AU2 development rush (meaning it wasn't taken in the second).

**Solution**: A simple bounty plugin. After discussion with the umpires, we're OK with the target's and bounty placer's names being public information. These do not use events unlike wantedness/etc. as there is no time expiry that makes sense to associate with a bounty. It's also a very rare occurrence and I'm not keen to put more questions on the umpire's to-do list whenever they add the event ("make the common case fast" and all that). Happy to hear thoughts on this. 

**Testing**: Manual. We really need to start using the testing framework. It's criminally underused.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -436,7 +436,7 @@ class CorePlugin(AbstractPlugin):
             SelectorList(
                 self.identifier + "_config",
                 title="Enable or disable plugins",
-                options=plugins,
+                options=sorted(plugins),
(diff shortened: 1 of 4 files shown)


thenewtonium (approved):
lgtm!

Issue #63, opened 2024-11-11
Assassin -> Resurrect as police


jsyiek:
We've had some issues with assassins turning into police (which is obviously something the umpires need to do) - it interacts badly with the TargetingPlugin which is almost completely stateless by design. To get this to work, we need a workaround.

This issue really consists of two parts:
1) Disable assassins turning normal assassins into police (done by @thenewtonium in 416cd076446d8a29c56b500590d34d2cc96f8627) 
2) Add a separate functionality to *clone* an assassin and make that assassin a police, and hide the original assassin (@thenewtonium has implemented the `hide` functionality as of https://github.com/jsyiek/AU2/pull/59)

PR #69, opened 2024-11-14
[63] Assassin -> Resurrect as Police

thenewtonium:
This builds on PR #59 to address issue #63.

It adds a new `Assassin -> Resurrect as Police` export to ~~`CorePlugin`~~ (EDIT: moved to `PolicePlugin`). The user selects from all *dead, non-police, non-hidden* assassins, and is prompted to give a new initial pseudonym for the selected assassin. The assassin is then cloned (with a new identifier), with the clone set to be police and have the entered pseudonym as its sole pseudonym.

### DIFF
--- a/AU2/database/model/Assassin.py
+++ b/AU2/database/model/Assassin.py
@@ -5,7 +5,7 @@ from typing import List, Dict, Any, Optional, Iterator, Union
 
 from AU2.database.GenericStateDatabase import GENERIC_STATE_DATABASE
 from AU2.database.model.PersistentFile import PersistentFile
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 
 from AU2 import TIMEZONE
 from AU2.plugins.util.date_utils import get_now_dt
@@ -61,6 +61,24 @@ class Assassin(PersistentFile):
         if not self.identifier:
             dotdotdot = "..." if len(self.pseudonyms[0]) > 15 else ""
             self.identifier = f"{self.real_name} ({self.pseudonyms[0][:15]}{dotdotdot}) ID: {self._secret_id}"
+
+    def clone(self, **changes):
+        """
+        Creates a clone of this assassin with a new identifier and specified changes.
+        This is used for resurrecting an assassin as police.
+
+        Args:
+            **changes: use keyword arguments to specify attributes of clone where they should differ from the original.
+                Setting a new identifier or _secret_id has no effect; the cloned assassin will still have these set
+                automatically.
+
+        Returns:
+            A clone of this Assassin, with a different identifier and _secret_id
+        """
+        # allow __post_init__ to generate the new identifier and _secret_id
+        changes["identifier"] = ""
+        changes["_secret_id"] = ""
+        return replace(self, **changes)
 
     def get_pseudonym(self, i: int) -> str:
         """
--- a/AU2/database/model/Assassin.py
+++ b/AU2/database/model/Assassin.py
@@ -42,7 +42,7 @@ class Assassin(PersistentFile):
     pseudonym_datetimes: Dict[int, dt.datetime] = field(
         default_factory=dict,
         metadata=config(
-            encoder=lambda d: {k: ts.timestamp() for k,ts in d.items()},
+            encoder=lambda d: {k: ts.timestamp() for k, ts in d.items()},
             decoder=lambda d: {int(k): dt.datetime.fromtimestamp(ts).astimezone().astimezone(TIMEZONE) for k,ts in d.items()}
         )
     )
--- a/AU2/database/model/Assassin.py
+++ b/AU2/database/model/Assassin.py
@@ -195,7 +195,7 @@ class Assassin(PersistentFile):
         Yields:
             the next pseudonym of the assassin that is valid before datetime `ts`.
         """
-        if not ts:
+        if ts is None:
             ts = get_now_dt()
         
         for (i, p) in enumerate(self.pseudonyms):
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -117,13 +117,6 @@ class CorePlugin(AbstractPlugin):
                 ((lambda: ASSASSINS_DATABASE.get_identifiers()),)
             ),
             Export(
-                "core_assassin_to_police",
-                "Assassin -> Resurrect as Police",
-                self.ask_core_plugin_assassin_to_police,
-                self.answer_core_plugin_assassin_to_police,
-                (self.gather_dead_non_police,)
-            ),
-            Export(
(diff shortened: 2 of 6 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -381,6 +406,27 @@ def answer_core_plugin_update_assassin(self, html_response_args: Dict):
             return_components += p.on_assassin_update(assassin, html_response_args)
         return return_components
 
+    def gather_dead_non_police(self) -> List[str]:
+        death_manager = DeathManager(perma_death=True)
```
jsyiek, line 410: I see why you've done it this way, but `perma_death=True` is incorrect behavior for May Week games. This is fine in some plugins (e.g., Competency and Targeting) that explicitly are not used for May Week. How about setting `perma_death=True` iff the TargetingPlugins is enabled?

AU2/plugins/CorePlugin.py
```diff
@@ -350,8 +377,6 @@ def answer_suppress_exports(self, htmlResponse):
 
     def ask_core_plugin_create_assassin(self):
         components = []
-        for p in PLUGINS:
```
jsyiek, line 353: Accidental delete?

AU2/database/model/Assassin.py
```diff
@@ -173,6 +195,9 @@ def pseudonyms_until(self, ts: dt.datetime = get_now_dt()) -> Iterator[str]:
         Yields:
             the next pseudonym of the assassin that is valid before datetime `ts`.
         """
+        if not ts:
```
jsyiek, line 198: We're checking for None here, so it's better to be explicit so that devs inputting non-`datetime` data by mistake don't accidentally get `dt.now()`

```suggestion
        if ts is None:
```

jsyiek (changes requested):
Initial review. I'm not sure what's from your previous PR and what's from this one - could you rebase master back into this branch to pick up changes?

AU2/plugins/CorePlugin.py
thenewtonium: Ah that's a good point.

It feels kind of hacky for `CorePlugin` to be checking for a specific plugin, but a better solution doesn't come to mind...

thenewtonium: Maybe since `perma_death=False` just makes the list empty (so effectively disables the export) the export itself should be part of `TargetingPlugin`?

AU2/plugins/CorePlugin.py
```diff
@@ -216,7 +225,6 @@ def on_assassin_request_update(self, assassin: Assassin):
             InputWithDropDown(self.html_ids["Water Status"], "Water Status", WATER_STATUSES, selected=assassin.water_status),
             InputWithDropDown(self.html_ids["College"], "College", COLLEGES, selected=assassin.college),
             LargeTextEntry(self.html_ids["Notes"], "Notes", default=assassin.notes),
-            #Checkbox(self.html_ids["Police"], "Police? (y/n)", checked=assassin.is_police)
```
NerdyNinja11, line 219: Maybe leave another comment explaining why this is commented, or just remove the comment altogether. Not obvious why it's there otherwise

AU2/plugins/CorePlugin.py
```diff
@@ -226,13 +234,12 @@ def on_assassin_update(self, assassin: Assassin, htmlResponse: Dict) -> List[HTM
...
         [assassin.edit_pseudonym(i, v.text, v.valid_from) for i, v in pseudonym_updates.edited.items()]
         [assassin.delete_pseudonym(i) for i in pseudonym_updates.deleted_indices]
-
+        # set other attributes
         assassin.real_name = htmlResponse[self.html_ids["Real Name"]]
         assassin.pronouns = htmlResponse[self.html_ids["Pronouns"]]
         assassin.address = htmlResponse[self.html_ids["Address"]]
         assassin.college = htmlResponse[self.html_ids["College"]]
         assassin.notes = htmlResponse[self.html_ids["Notes"]]
-        #assassin.is_police = htmlResponse[self.html_ids["Police"]]
```
NerdyNinja11, line 235: (same as above) Maybe leave another comment explaining why this is commented, or just remove the comment altogether. Not obvious why it's there otherwise

NerdyNinja11 (commented):
Not sure how useful it would be, but since we're in the business now of adding tests, you could add a test for this?

AU2/plugins/CorePlugin.py
thenewtonium: (The change here *is* removing the comment (that's what the - and the red mean).)

NerdyNinja11: Oh oops

AU2/plugins/CorePlugin.py
```diff
@@ -401,6 +408,29 @@ def answer_core_plugin_update_assassin(self, html_response_args: Dict):
             return_components += p.on_assassin_update(assassin, html_response_args)
         return return_components
 
+    def gather_dead_non_police(self) -> List[str]:
+        # use targeting plugin as proxy for whether we have perma-death
```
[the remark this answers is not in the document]
NerdyNinja11: There are a few edge cases if the umpire decides to run their game in an unusual way, but I think that would work.
Using CompetencyPlugin seems fine though (although I do agree that it seems kinda hacky)

AU2/test/plugins/util/custom_plugins/test_PolicePlugin.py
```diff
@@ -0,0 +1,73 @@
...
+        # check that the list of players that can be resurrected is correct
+        # (only the dead non-police, non-hidden assassins should be shown!)
+        assert set(plugin.gather_dead_non_police()) == {p[i] + " identifier" for i in range(k, 2*k-1)}
+
+        # check that players get resurrected correctly
+        ##############################################
+        new_pseudonym = "New Pseudonym"
+        old_assassin = ASSASSINS_DATABASE.get(p[k] + " identifier")
+        old_idents = set(ASSASSINS_DATABASE.get_identifiers(include_hidden=lambda _: True))
+        # the way identifiers are generated in test mode messes with cloning
```
NerdyNinja11, line 43: Since this has caused issues for both of us, maybe we need to rewrite how test mode treats assassins a bit.

NerdyNinja11 (approved):
lgtm

AU2/test/plugins/util/custom_plugins/test_PolicePlugin.py
thenewtonium: Yeah...
In this particular case the annoying thing is that if identifiers worked normally in test mode all the other tests would need to be rewritten. Luckily we don't have that many tests at the moment...

Issue #64, opened 2024-11-11
PageGeneratorPlugin replaces [PX] specifiers even if player not in event


jsyiek:
Umpires occasionally want to include lots of players in an event. It can be inconvenient to do so (especially if need to update later) and so may miss adding people sometimes. This can result in hanging `[PX]` specifiers in the headline/reports which are difficult to debug. It would be nice for the umpires to be able to add people to the report/headline without adding them to the event.

PR #70, opened 2024-11-14
[64] Allow pseudonym replacement even if assassin is not in event

jsyiek:
Implements #64 

**Problem**: By default, when [PX]/etc. is in a report/headline and is not an assassin explicitly added to an event, the pseudonym is not rendered.

**Solution**: Allow it to be rendered by performing some regex search to figure out which assassins are in the report.

**Testing**: Manual

### DIFF
--- a/AU2/plugins/custom_plugins/PageGeneratorPlugin.py
+++ b/AU2/plugins/custom_plugins/PageGeneratorPlugin.py
@@ -1,6 +1,9 @@
 import datetime
 import os
+import re
 import zlib
+
+
 from typing import List
 
 from AU2 import ROOT_DIR
@@ -96,6 +99,8 @@ HARDCODED_COLORS = {
     "Valheru": "#03FCEC",
     "Vendetta": "#B920DB"
 }
+
+FORMAT_SPECIFIER_REGEX = r"\[[D,P,N]([0-9]+)(?:_([0-9]+))?\]"
 
 
 def weeks_and_days_to_str(start: datetime.datetime, week: int, day: int) -> str:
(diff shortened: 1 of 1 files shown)


thenewtonium (changes requested):

thenewtonium:
Testing this it has occured to me that hiding deleted pseudonyms makes it hard to figure out what the appropriate index for a given pseudonym is when using the [PX_i] format.

Perhaps blank pseudonyms shouldn't be hidden in Assassin -> Update (but still be hidden in Event -> Create and Event -> Update and when rendering [DX])...

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -227,6 +232,38 @@ def on_page_generate(self, _) -> List[HTMLComponent]:
...
+
+                    color = get_color(
+                        pseudonym,
+                        dead=death_manager.is_dead(assassin_model),
+                        incompetent=competency_manager.is_inco_at(assassin_model, e.datetime),
+                        is_police=assassin_model.is_police,
+                        is_wanted=wanted_manager.is_player_wanted(assassin_model.identifier, time=e.datetime)
+                    )
+
+                    candidate_pseudonyms.append((assassin_model, pseudonym, color))
```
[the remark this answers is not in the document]
jsyiek: Ah, good spot. The problem looks like it comes from here:

```
pseudonym_index = int(match[0] or 0)
```

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -227,6 +232,38 @@ def on_page_generate(self, _) -> List[HTMLComponent]:
...
```
thenewtonium, line 265: This still needs to be moved after the iteration over `e.assassins.items()`

thenewtonium (changes requested):

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
thenewtonium: (not line +235 ofc)

thenewtonium (approved):
Lgtm!

PR #71, opened 2024-11-15
Gigabolt Option

### LINKED ISSUE #21: GigaboltPlugin
Modern games use a 'gigabolt', whereby all players who have never attempted or participated in the game are thunderbolted a few weeks after incobash. This has been done manually and painfully by the umpires. I'd like a plugin that automates finding and eliminating these inactive players.

**How it should be done**: Collect all players who have participated in any event, and remove them from a list of all assassins. Those left are the list of players to be thunderbolted. Give the plugin an `export` option to perform this action (see the architecture of other plugins in AU2 for how the export/ask/answer system works).

You need to stress test any Gigabolt implementation against the targeting algorithm. Dynamic updates to the targeting graph require solving a constraint satisfaction problem which are notoriously intractable. My implementation uses an attempt-bounded brute force (bounded exponentially by the number of deaths, which are usually very small), which obviously would be insufficient for large numbers of updates.

NerdyNinja11:
Adds functionality to close #21 

Adds a `Export` in `CompetencyPlugin` which asks for:
A list of assassins to thundebolt (Pre-selected to be those with no kills and no attempts)
A member of the police (i.e. the umpire)
A date
A headline, with the option to be blank, and also the ability to include the number of gigabolted players in it.

It will then subdivide the players (so the targeting plugin can handle it), and create events until all the selected players have been killed.
If the headline is blank, all events will be hidden; otherwise, the first will be unhidden.

Testing:
I've tested with multiple `RandomGame`s, and `TargetingPlugin` handles it fine (obviously as long as there are enough players left)

### DIFF
--- a/AU2/plugins/custom_plugins/CompetencyPlugin.py
+++ b/AU2/plugins/custom_plugins/CompetencyPlugin.py
@@ -13,8 +13,10 @@ from AU2.html_components.DependentComponents.AssassinDependentSelector import As
 from AU2.html_components.SimpleComponents.DatetimeEntry import DatetimeEntry
 from AU2.html_components.MetaComponents.Dependency import Dependency
 from AU2.html_components.SimpleComponents.InputWithDropDown import InputWithDropDown
+from AU2.html_components.SimpleComponents.LargeTextEntry import LargeTextEntry
 from AU2.html_components.SimpleComponents.IntegerEntry import IntegerEntry
 from AU2.html_components.SimpleComponents.Label import Label
+from AU2.html_components.SimpleComponents.SelectorList import SelectorList
 from AU2.plugins.AbstractPlugin import AbstractPlugin, ConfigExport, Export
 from AU2.plugins.CorePlugin import registered_plugin
 from AU2.plugins.constants import WEBPAGE_WRITE_LOCATION
@@ -55,6 +57,14 @@ DEAD_INCOS_TABLE_ROW_TEMPLATE = """
 
 NO_INCOS = """<p xmlns="">Hmm, how interesting... no one is incompetent yet.</p>"""
 
+DEFAULT_GIGABOLT_HEADLINE = """# Write a headline for the gigabolt event
+# Use [num_players] to specify the number of players killed
+# Leave this blank for a fully hidden event
+# Lines prefixed with # will be ignored
+# 
+[num_players] assassins are eliminated for inactivity!
+"""
+
 INCOS_PAGE_TEMPLATE: str
 with open(os.path.join(ROOT_DIR, "plugins", "custom_plugins", "html_templates", "inco.html"), "r", encoding="utf-8", errors="ignore") as F:
     INCOS_PAGE_TEMPLATE = F.read()
@@ -75,7 +85,10 @@ class CompetencyPlugin(AbstractPlugin):
             "Competency": self.identifier + "_competency",
             "Datetime": self.identifier + "_datetime",
             "Auto Competency": self.identifier + "_auto_competency",
-            "Attempts": self.identifier + "_attempts"
+            "Attempts": self.identifier + "_attempts",
+            "Gigabolt": self.identifier + "_gigabolt",
+            "Headline": self.identifier + "_gigabolt_headline",
+            "Umpire": self.identifier + "_umpire"
         }
 
         self.plugin_state = {
@@ -94,6 +107,12 @@ class CompetencyPlugin(AbstractPlugin):
                 display_name="Competency -> Show deadlines",
                 ask=self.ask_show_inco_deadlines,
                 answer=self.answer_show_inco_deadlines
+            ),
+            Export(
+                identifier="competency_plugin_gigabolt",
+                display_name="Competency -> Gigabolt",
+                ask=self.gigabolt_ask,
+                answer=self.gigabolt_answer
             )
         ]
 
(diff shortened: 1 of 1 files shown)


jsyiek:
I haven't had a chance to give a proper review yet, but I don't see any automated tests here. Could you write several please? We need this functionality to be completely bug-free on first use in prod.

jsyiek:
E.g., several that have assassins do varying levels of actions and create events with appropriate attempt qualifiers, and then check that no assassins who did attempts/kills get gigabolted.

thenewtonium (commented):
Aside from what Jamie said, everything looks ok to me (but a couple minor comments)

NerdyNinja11:
Tests added (never done tests before, so probably wants some critique/tweaking)

NerdyNinja11:
Multiple attempt support added, still compatible with existing database.

AU2/test/plugins/util/custom_plugins/test_CompetencyPlugin.py
```diff
@@ -75,18 +76,127 @@ def test_competency_plugin_event_create(self):
...
+
+        assert len(incos) == 2
+
+    @plugin_test
+    def test_gigabolt(self):
+        """
+        Test that only non-police players with no kills and no attempts will be eliminated.
+        The dangerous logic for gigabolt is in CompetencyPlugin.gigabolt_ask (i.e. the default selections),
+        while this only tests gigabolt_answer directly. It is mostly just code copy-pasted from gigabolt_ask,
+        and thus is not really a good test
```
thenewtonium, line 161: You could perhaps call gigabolt_ask and then extract the list from the `defaults` value of the relevant component?
(sth like `next((c.defaults for c in plugin.gigabolt_ask() if c.identifier == plugin.html_ids["Gigabolt"]))` would do it in one line) then pass this into `gigabolt_answer`?

AU2/test/plugins/util/custom_plugins/test_CompetencyPlugin.py
```diff
@@ -75,18 +76,127 @@ def test_competency_plugin_event_create(self):
...
+                .assassin(p[19]).is_police()
+                .assassin(p[18]).is_police()
+                .assassin(p[0]).kills(p[1])
+                .assassin(p[2]).kills(p[3])
+                .assassin(p[4]).kills(p[5])
+                .assassin(p[6]).kills(p[7])
+                .assassin(p[15]).is_involved_in_event(
+            pluginState={"CompetencyPlugin": {"attempts": [p[15] + " identifier"]}})
+                .assassin(p[16]).is_involved_in_event(
+            pluginState={"CompetencyPlugin": {"attempts": [p[15] + " identifier"] * 3}}))
```
thenewtonium, line 174: Is this meant to be
```suggestion
                .assassin(p[16]).is_involved_in_event(
            pluginState={"CompetencyPlugin": {"attempts": [p[16] + " identifier"] * 3}}))
```
As is the test actually fails because player 16 hasn't actually made any attempts or kills, so there are only 7 live players after gigabolting!

thenewtonium (changes requested):
See comments, but personally I am satisfied that gigabolt does work correctly.
I've manually tested against the current game database and none of the pseudonyms of suggested players to thunderbolt appear in any of the headlines, and after fixing the test case (the event that is meant to give player 16 3 attempts actually gives them to player 15) it passed that as well.

AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -232,7 +322,12 @@ def on_event_create(self, e: Event, htmlResponse) -> List[HTMLComponent]:
         if self.html_ids["Competency"] in htmlResponse:
             e.pluginState.setdefault(self.identifier, {})[self.plugin_state["COMPETENCY"]] = htmlResponse[self.html_ids["Competency"]]
         if self.html_ids["Attempts"] in htmlResponse:
-            e.pluginState.setdefault(self.identifier, {})[self.plugin_state["ATTEMPTS"]] = htmlResponse[self.html_ids["Attempts"]]
+            # This currently stores attempts as ['a', 'a'] for two attempts by player a
```
thenewtonium, line 325: On this: for storing duplicates of the same id in `attempts` to actually work for granting competency you need to modify competency manager as well; currently it does
```python3
            for assassin_id in e.pluginState.get("CompetencyPlugin", {}).get("attempts", []):
                # Logic for not increasing attempts if player got a player kill in same event.
                for (killer, victim) in e.kills:
                    if killer == assassin_id and not ASSASSINS_DATABASE.get(victim).is_police:
                        break
                else:
                    self.attempts_since_kill[assassin_id] += 1
```
which will only count one attempt per event no matter the number of duplicates!

NerdyNinja11: Pretty sure it does work with no changes - it will increment the counter for each attempt and extend competency if ever a multiple of two.
Besides, both the test and manual testing suggests that it works

AU2/test/plugins/util/custom_plugins/test_CompetencyPlugin.py
NerdyNinja11: Not sure how I didn't catch that...

NerdyNinja11:
> none of the pseudonyms of suggested players to thunderbolt appear in any of the headlines

That's surprising, I'd expect some of the incos who've been attempted on in headlines to be thunderbolted. Unless there have been loads of hidden attempts this game, which the umpires decided to allow

AU2/plugins/custom_plugins/CompetencyPlugin.py
thenewtonium: Oh you're right actually
I think I must have misread the first `for` as an `if` or something  😶

thenewtonium:
> That's surprising, I'd expect some of the incos who've been attempted on in headlines to be thunderbolted. Unless there have been loads of hidden attempts this game, which the umpires decided to allow
It turns out I did miss one inco who had been attempted on and was (correctly) suggested for gigabolting. I think it's reasonable to assume that most unsuccessful attempts on incos have been ones where the inco is unaware of the attempt, so will probably be hidden.

PR #76, opened 2024-11-18
[62] Improved Competency -> Show deadlines [REQUIRES `tabulate`]

thenewtonium:
The user can now search for assassins, and the results are given using a new `Table` HTMLcomponent which is rendered by `inquirer_cli` using [`tabulate`](https://pypi.org/project/tabulate/).

### DIFF
--- a/AU2/plugins/custom_plugins/CompetencyPlugin.py
+++ b/AU2/plugins/custom_plugins/CompetencyPlugin.py
@@ -17,6 +17,7 @@ from AU2.html_components.SimpleComponents.LargeTextEntry import LargeTextEntry
 from AU2.html_components.SimpleComponents.IntegerEntry import IntegerEntry
 from AU2.html_components.SimpleComponents.Label import Label
 from AU2.html_components.SimpleComponents.SelectorList import SelectorList
+from AU2.html_components.SimpleComponents.NamedSmallTextbox import NamedSmallTextbox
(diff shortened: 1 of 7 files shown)


AU2/plugins/custom_plugins/CompetencyPlugin.py
```diff
@@ -299,23 +307,32 @@ def answer_show_inco_deadlines(self, _):
...
-            labels.append(Label(f"{a} -> {datetime.datetime.strftime(d, DATETIME_FORMAT)}"))
-            labels.append(Label("__________________________________________________________________"))
-        return labels
+        now = get_now_dt()
+        deadlines = []
+        for a in ASSASSINS_DATABASE.get_filtered(
+                include=lambda x: any(t.lower() in x.identifier.lower() for t in search_terms)
+        ):
+            d = competency_manager.get_deadline_for(a)
+            datetime_str = datetime.datetime.strftime(d, DATETIME_FORMAT) if d else "n/a"
```
NerdyNinja11, line 318: It would be useful if it gave some more info than just n/a. I've never had a good look at exactly how competency is managed in AU2, but the n/a s just don't tell me anything to aid debugging (as in, I wouldn't expect them ever, except on maybe police)

thenewtonium: That's a good point actually, it's odd that any assassins end up with n/a at all.

NerdyNinja11 (approved):
lgtm

PR #77, opened 2024-11-20
[53] notepad via pyinquirer on windows uses utf-8 by default

### LINKED ISSUE #53: Emojis don't work in reports (Possibly only on Windows)
Steps to reproduce:

- Paste an emoji into the notepad document for a report in an event
- Generate pages for that event
- Emoji replaced by 4 other characters on webpage

The issue is very early in the pipeline, i.e. the emoji in the returned `report` in inquirer_cli, 
`report = inquirer_prompt_with_abort(q)["report"]` 
has been converted to 4 characters, and it is stored in the json DB in that form.
Likely also an issue with headlines, but untested

NerdyNinja11:
fixes #53, after way too much trial and error

### DIFF
--- a/AU2/__init__.py
+++ b/AU2/__init__.py
@@ -4,5 +4,14 @@ import pytz
 
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 BASE_WRITE_LOCATION = os.path.expanduser("~/database")
-
 TIMEZONE = pytz.timezone("Europe/London")
+
+# Sets the encoding to utf-8 for pyinquirer on Windows machines, since that isn't the default for some reason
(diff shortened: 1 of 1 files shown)


NerdyNinja11:
I'm really hoping this isn't one of those things that is only fixed on my machine

thenewtonium:
Unfortunately I can't really test this because I don't have access to a Windows machine. It doesn't look like it would do any harm at least though. But then again `notepad /W` causing a crash was a surprise ¯\\\_(ツ)\_/¯

jsyiek (approved):
i hereby certify this free of nasal demons

PR #78, opened 2024-11-20
[75] Added PgUp, PgDown, Home and End key support

thenewtonium:
This is done by using a wrapper function which adds extra key handling to `inquirer.render.console._list.List.process_input` and `inquirer.render.console._checkbox.Checkbox.process_input`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,3 +167,4 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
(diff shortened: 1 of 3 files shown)


NerdyNinja11 (approved):
I think I'd prefer if it went a full page down instead of just half_options - but I see that it behaves weirdly otherwise since the selected option is always in the centre.
If that isn't trivial to adjust, the rest looks good - just merge it

thenewtonium:
Looking at the source code it doesn't like like it would be *too* difficult to change how options are displayed to not centre the selected option, but I'd rather not make to much of a habit of messing with external modules in this way...

Issue #55, opened 2024-11-08
Search for assassins


jsyiek:
Title. When adding/updating/etc. an assassin

PR #83, opened 2024-11-28
Implemented `DangerousConfigExport`

thenewtonium:
**Problem:** Certain config options (mainly to do with targeting) break the game if changed after a game has started. It would be useful to communicate this to users.

**Solution:** Have a subclass of `ConfigExport` called `DangerousConfigExport`, and have these coloured red in `Plugin config -> Plugin-specific parameters`. The config options I have classed as being dangerous are
- `Competency -> Change Auto Competency`
- `Targeting Graph -> Set player seeds`
- `Targeting Graph -> Set random seed`

**Remarks:**` gather_config_names` had to be reworked since otherwise the ANSI codes messed with the ordering of the config options. The module`colorama` is used in `inquirer_cli` to ensure the ANSI codes are rendered correctly on Windows (this is only an issue for older terminals), but not having it installed will only cause a warning rather than a crash.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -8,6 +8,7 @@ from AU2.database.GenericStateDatabase import GENERIC_STATE_DATABASE
 from AU2.database.model import Assassin, Event
 from AU2.html_components import HTMLComponent
 from AU2.html_components.SpecialComponents.EditablePseudonymList import EditablePseudonymList, PseudonymData
+from AU2.html_components.SpecialComponents.ConfigOptionsList import ConfigOptionsList
 from AU2.html_components.DependentComponents.AssassinDependentReportEntry import AssassinDependentReportEntry
 from AU2.html_components.DependentComponents.AssassinPseudonymPair import AssassinPseudonymPair
 from AU2.html_components.SimpleComponents.Checkbox import Checkbox
@@ -153,7 +154,7 @@ class CorePlugin(AbstractPlugin):
                 "Plugin config -> Plugin-specific parameters",
                 self.ask_config,
                 self.answer_config,
-                (self.gather_config_names,)
+                (self.gather_config_options,)
             )
         ]
 
@@ -172,7 +173,7 @@ class CorePlugin(AbstractPlugin):
         ]
 
         self.config_exports = [
-            ConfigExport(
+            DangerousConfigExport(
                 "core_plugin_set_game_start",
                 "CorePlugin -> Set game start",
                 self.ask_set_game_start,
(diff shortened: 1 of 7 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -523,10 +523,16 @@ def gather_config_names(self):
...
-                names.append(c.display_name)
-        names.sort()
+        configs = (c for p in PLUGINS for c in p.config_exports)
+        for c in sorted(configs, key=lambda e: e.display_name):
+            disp = c.display_name
+            # display dangerous config options in red if the game has already started
+            if isinstance(c, DangerousConfigExport) and get_game_start() <= get_now_dt():
+                # this is kind of cheating because it's not really html compatible
+                # a `ansi2html` library does exist though,
+                # or we can create a specific HTMLcomponent to implement the colouring
```
jsyiek, line 533: Let's make this an HTML component with a 'color' attribute (which is mapped at run-time from a string - e.g. "red" - to an implementation of red)

AU2/plugins/CorePlugin.py
```diff
@@ -523,10 +523,16 @@ def gather_config_names(self):
...
         names = []
-        for p in PLUGINS:
-            for c in p.config_exports:
-                names.append(c.display_name)
-        names.sort()
+        configs = (c for p in PLUGINS for c in p.config_exports)
+        for c in sorted(configs, key=lambda e: e.display_name):
+            disp = c.display_name
+            # display dangerous config options in red if the game has already started
+            if isinstance(c, DangerousConfigExport) and get_game_start() <= get_now_dt():
```
jsyiek, line 530: `DangerousConfigExport`s should probably make themselves a little annoying to modify after game start. Could we implement a text challenge for all `DangerousConfigExport`s, similar to what is done for `Reset Database` in (iirc) `LocalBackupPlugin`?  You might want to move the challenge into a library location to reuse the functionality across both.

jsyiek (changes requested):
A few suggestions for improvements

AU2/plugins/CorePlugin.py
thenewtonium: I've just remembered why I did it like this in the first place[^1]. The list of plugin options is given using an `options_functions` rather than returning a `HTMLcomponent`. So having a HTMLcomponent would require a reworking of the plugin framework (say by having the `options_functions` return  HTMLcomponents rather than a list of options).
I think this would be reasonable to do (it would also allow us to extend #55 to `Assassin -> Update` for example), since I could have it check if the `options_functions` for a given export does return a HTMLcomponent and if not fall back to the current behaviour to avoid having to modify every single existing plugin export.

[^1]: (I started writing this code a couple days ago)

AU2/plugins/CorePlugin.py
thenewtonium: I think for this I would want to combine what I wrote above with implementing what you suggested for #57 (letting each options function take the results of all the previous options functions) so that the second options function can do the challenge before `config_ask` is called, in the case that a `DangerousConfigExport` is selected.

AU2/plugins/CorePlugin.py
jsyiek: Or, you add a meta component to implement the challenge, and then use `@property` to override accesses to underlying element and wrap it in the new meta component

jsyiek: A simple improvement would be to return a list of `Options` for some new class `Option` which has associated metadata (such as color) and then create a "ColoredSelectorList". Idk. 

Or, have the front-end tell you whether it is HTML or not? Maybe you could find a good place for front-end to tell back-end its name.

jsyiek: I'm also happy with the proposed changes you made (and would prefer those, if you have time)

jsyiek (approved):

Issue #80, opened 2024-11-26
Sanity check plugin


jsyiek:
Extensible framework for auto-amending pages etc. for common mistakes

E.g., [38] -> [P38]

PR #84, opened 2024-11-28
[80] Sanity checking framework

jsyiek:
Implements #80 

**Problem**: There are some usability issues with AU2 - to name a few, umpires occasionally mistype [P38] as [38] or forget to put `<!--HTML-->` in front of an HTML report. There's not a good error-checking facility in the tool to pick up on these unavoidable occasional mistakes, leading to these bleeding into production webpages.

**Solution**: I've implemented an extensible sanity checking framework - only requiring two functions to be implemented per sanity check. I've implemented one check for now as a proof of concept - hopefully we can work together to add more in future.

This was not trivial to do, because we really need the following requirements:
- Must not introduce more exports (an error-checking facility you can forget to click isn't useful)
- Must not be its own plugin (an error-checking facility you can forget to turn on isn't useful)
- Must run when you generate a page (this is a key point where errors become visible to players)
- Must be HTML-compatible (in case we ever make this front-end - so no dictionaries, tuples, etc.)
- Must clean up all reports BEFORE any pages generate (so must have special hoist status to the top of `answer_generate_pages`). Currently, execution order is undefined (NASAL DEMONS!)

With these in mind, why I did it how I did it hopefully makes sense.

### DIFF
--- a/AU2/plugins/sanity_checks/__init__.py
+++ b/AU2/plugins/sanity_checks/__init__.py
@@ -1,0 +1,8 @@
+from typing import Dict
+
+from AU2.plugins.sanity_checks.MissingHtmlSpecifier import MissingHtmlSpecifier
+from AU2.plugins.sanity_checks.model.SanityCheck import SanityCheck
+
+SANITY_CHECKS: Dict[str, SanityCheck] = {
+    "Missing_HTML_Specifiers": MissingHtmlSpecifier()
+}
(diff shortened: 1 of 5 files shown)


AU2/plugins/sanity_checks/__init__.py
```diff
@@ -0,0 +1,8 @@
+from typing import Dict
+
+from AU2.plugins.sanity_checks.MissingHtmlSpecifier import MissingHtmlSpecifier
+from AU2.plugins.sanity_checks.lib.SanityCheck import SanityCheck
```
thenewtonium, line 4: It seems like you forgot to include `AU2.plugins.sanity_checks.lib` in the PR.

thenewtonium (changes requested):
There seems to be code missing from this PR... (specifically `sanity_checks/lib`)
So I can't test it. But nothing jumps out at me from reading the code.

Issue #90, opened 2024-11-30
Sanity check - [38] instead of [P38]


jsyiek:
Sometimes umpires forget to put the P in front of a pseudonym, and this can be corrected with a sanity check

PR #96, opened 2024-11-30
[90] Sanity check - [X] instead of [PX]

jsyiek:
Implements #90 

**Problem**: Umpires sometimes put [X] instead of [PX]. 

**Solution**: This sanity check asks the umpires if they want to replace [X] with [PX]. Currently, the only instances in prod of this are in Peter and Emma's reports, and both are meant to represent references - so I'll need to let the umpires know of the change and make sure they don't choose to replace them!

### DIFF
--- a/AU2/plugins/sanity_checks/IncorrectPseudonymFormatter.py
+++ b/AU2/plugins/sanity_checks/IncorrectPseudonymFormatter.py
@@ -6,14 +6,20 @@ from AU2.html_components import HTMLComponent
 from AU2.plugins.sanity_checks.model.SanityCheck import SanityCheck
 from AU2.plugins.sanity_checks.model.SanityCheck import Suggestion
 
+
 class IncorrectPseudonymFormatter(SanityCheck):
+    """
+    Finds and detects cases where the umpire writes pseudonysm as
+    [X] instead of [PX] and corrects them.
+    """
 
     identifier = "Incorrect_Pseudonym_Formatter"
 
     def _fix(self, string: str, fixes: Dict[str, str]):
-        search_pattern = r"\[([0-9]+)\]"
+        # Matches any number of white space, one word, then [X], then one word and space
+        search_pattern = r"(\S*[^\w\n]*)\[([0-9]+)\]([^\w\n]*\S*)"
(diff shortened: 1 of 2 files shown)


NerdyNinja11:
I wasn't really following the sanity check framework code, but this definitely needs comments explaining what is is checking exactly. Also a comment explaining the regex would be good, since mere mortals cannot simply read and understand regex.

AU2/plugins/sanity_checks/IncorrectPseudonymFormatter.py
```diff
@@ -0,0 +1,43 @@
...
+from AU2.plugins.sanity_checks.model.SanityCheck import Suggestion
+
+class IncorrectPseudonymFormatter(SanityCheck):
+
+    identifier = "Incorrect_Pseudonym_Formatter"
+
+    def _fix(self, string: str, fixes: Dict[str, str]):
+        search_pattern = r"\[([0-9]+)\]"
+        for match in re.findall(search_pattern, string):
+            fixes[f"[{match}]"] = f"[P{match}]"
```
thenewtonium, line 16: Imo the way the suggestions are presented isn't very enlightening, since it doesn't show the context where the mistake appears. This could be fixed by including some surrounding text, e.g. using the regex
```regex
(\S*[^\w\n]*)\[([0-9]+)\]([^\w\n]*\S*)
```
(which includes the neighbouring words)

There is an edge case where that doesn't work correctly, namely when we have overlapping matches. But at the very least it raises awareness of the nearby error ¯\\\_(ツ)\_/¯.

```suggestion
    def _fix(self, string: str, fixes: Dict[str, str]):
        search_pattern = r"(\S*[^\w\n]*)\[([0-9]+)\]([^\w\n]*\S*)"
        for match in re.findall(search_pattern, string):
            fixes[f"{match[0]}[{match[1]}]{match[2]}"] = f"{match[0]}[P{match[1]}]{match[2]}"

```

jsyiek: Good idea! I'll use a variant of this.

thenewtonium (approved):
Lgtm

PR #97, opened 2024-12-01
Remove `colorama` warning

thenewtonium:
As it is unnecessary now that colouring is not the only safeguard against accessing DangerousConfigExports after game start.

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -737,7 +737,7 @@ def main():
     try:
         from colorama import just_fix_windows_console
         just_fix_windows_console()
-    except Exception:
+    except:
         pass
 
     while True:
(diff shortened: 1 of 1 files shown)


AU2/frontends/inquirer_cli.py
```diff
@@ -732,15 +732,13 @@ def replace_overrides(component_list: List[HTMLComponent], existing_overrides={}
...
+    # this fixes rendering of ANSI codes in old Windows terminals
+    # if it fails we don't want it to cause a crash though
     try:
         from colorama import just_fix_windows_console
         just_fix_windows_console()
-    except ModuleNotFoundError:
-        if os.name == "nt":
-            print("[WARNING] `colorama` is not installed; if you are using an old terminal, "
-                  "colours may not render correctly.")
+    except Exception:
```
jsyiek, line 740:
```suggestion
    except:
```

jsyiek (approved):
Fix nit, squash commits, then lgtm

PR #100, opened 2025-01-07
[73] Stats Page

thenewtonium:
This PR adds an export `Scoring -> Generate stats page` to `ScoringPlugin`.
The stats page is configurable, allowing the user to select which columns should be included. In addition to the standard columns
- Real Name
- Pseudonym
- Number of Kills
- Conkers Score

there is the option to include
- Number of Attempts
- Score (as calculated for Open Season)
- Died at (which includes a link to the relevant event)
- Position (which is just where the player appears in the ordering used on the stats page)

The user can choose between two ordering options: by number of kills (the way recent stats pages have been ordered), or based on order of death (the way stats pages were ordered previously).

Furthermore there is an option to generate a visualisation of the kill graph using `pyvis`. This is generated as a separate page which is linked to from the stats page.

There's still a bunch of scope for more game stats as @NerdyNinja11 has suggested, but this is sufficient for mimicking the stats pages of previous games at least.

### DIFF
--- a/AU2/plugins/custom_plugins/PageGeneratorPlugin.py
+++ b/AU2/plugins/custom_plugins/PageGeneratorPlugin.py
@@ -2,9 +2,10 @@ import datetime
 import os
 import re
 import zlib
+import itertools
 
-
-from typing import List
+from typing import List, Tuple, Optional
+from collections import namedtuple
 
 from AU2 import ROOT_DIR
 from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
@@ -82,9 +83,18 @@ WANTED_COLS = [
     "#ff0033", "#cc3333", "#ff3300"
 ]
 
+
+def event_url(e: Event) -> str:
+    """
+    Generates the (relative) url pointing to this event's appearance on the news pages
+    """
+    week = date_to_weeks_and_days(get_game_start().date(), e.datetime.date()).week
+    return f"news{week:02}.html#{e._Event__secret_id}"
+
+
 HEAD_HEADLINE_TEMPLATE = """
     <div xmlns="" class="event">
-  [<a href="news{NUMBER}.html#{ID}">{TIME}</a>]
+  [<a href="{URL}">{TIME}</a>]
    <span class="headline">{HEADLINE}</span><br/></div>"""
 
 HEAD_DAY_TEMPLATE = """<h3 xmlns="">{DATE}</h3> {HEADLINES}"""
@@ -101,6 +111,22 @@ HARDCODED_COLORS = {
 }
 
 FORMAT_SPECIFIER_REGEX = r"\[[D,P,N]([0-9]+)(?:_([0-9]+))?\]"
+
+WeekAndDay = namedtuple("WeekAndDay", ["days_since_start", "week", "day_of_week"])
+def date_to_weeks_and_days(start: datetime.date, date: datetime.date) -> WeekAndDay:
+    """
+    Converts param `date` into a namedtuple `WeekAndDay` of:
+        - days_since_start: number of days since param `start`
+        - week: week number (week starting on param `start` is week 1)
+        - day_of_week: day number in week (taking values 0 to 6 inclusive)
+    """
+    days_since_start = (date - start).days
+    week = days_since_start // 7 + 1
+    day = days_since_start % 7
+    if week < 0:
+        week = 0
+        day = days_since_start + 7
+    return WeekAndDay(days_since_start, week, day)
 
 
 def weeks_and_days_to_str(start: datetime.datetime, week: int, day: int) -> str:
@@ -154,6 +180,55 @@ def substitute_pseudonyms(string: str, main_pseudonym: str, assassin: Assassin, 
     string = string.replace(f"[N{id_}]",
                             PSEUDONYM_TEMPLATE.format(COLOR=color, PSEUDONYM=soft_escape(assassin.real_name)))
     return string
+
+
+def render_headline_and_reports(e: Event,
+                                death_manager: DeathManager,
+                                competency_manager: CompetencyManager,
+                                wanted_manager: WantedManager) -> Tuple[str, List[str]]:
+    headline = e.headline
+
+    reports = {(playerID, pseudonymID): soft_escape(report) for (playerID, pseudonymID, report) in
+               e.reports}
+
+    candidate_pseudonyms = []
+
+    texts_to_search = [r for r in itertools.chain((headline,), reports.values())]
+
+    for r in texts_to_search:
+        for match in re.findall(FORMAT_SPECIFIER_REGEX, r):
+            assassin_secret_id = int(match[0])
+
+            assassin_model: Assassin
+            for a in ASSASSINS_DATABASE.assassins.values():
+                if int(a._secret_id) == assassin_secret_id:
+                    assassin_model = a
+                    break
+            else:
+                continue
+
+            if any(c[0].identifier == assassin_model.identifier for c in candidate_pseudonyms):
+                continue
+
+            pseudonym_index = int(match[1] or e.assassins.get(assassin_model.identifier, 0))
+            pseudonym = assassin_model.get_pseudonym(pseudonym_index)
+
+            color = get_color(
+                pseudonym,
+                dead=death_manager.is_dead(assassin_model),
+                incompetent=competency_manager.is_inco_at(assassin_model, e.datetime),
+                is_police=assassin_model.is_police,
+                is_wanted=wanted_manager.is_player_wanted(assassin_model.identifier, time=e.datetime)
+            )
+
+            candidate_pseudonyms.append((assassin_model, pseudonym, color))
+
+    for (assassin_model, pseudonym, color) in candidate_pseudonyms:
+        headline = substitute_pseudonyms(headline, pseudonym, assassin_model, color, e.datetime)
+        for (k, r) in reports.items():
+            reports[k] = substitute_pseudonyms(r, pseudonym, assassin_model, color, e.datetime)
+
+    return headline, reports
 
 
 @registered_plugin
@@ -228,46 +303,12 @@ class PageGeneratorPlugin(AbstractPlugin):
             if e.pluginState.get(self.identifier, {}).get(self.plugin_state["HIDDEN"], False):
                 continue
 
-            headline = e.headline
-
-            reports = {(playerID, pseudonymID): soft_escape(report) for (playerID, pseudonymID, report) in e.reports}
-
-            candidate_pseudonyms = []
-
-            texts_to_search = [r for r in reports.values()] + [headline]
-
-            for r in texts_to_search:
-                for match in re.findall(FORMAT_SPECIFIER_REGEX, r):
-                    assassin_secret_id = int(match[0])
-
-                    assassin_model: Assassin
-                    for a in ASSASSINS_DATABASE.assassins.values():
-                        if int(a._secret_id) == assassin_secret_id:
-                            assassin_model = a
-                            break
-                    else:
-                        continue
-
-                    if any(c[0].identifier == assassin_model.identifier for c in candidate_pseudonyms):
-                        continue
-
-                    pseudonym_index = int(match[1] or e.assassins.get(assassin_model.identifier, 0))
-                    pseudonym = assassin_model.get_pseudonym(pseudonym_index)
-
-                    color = get_color(
-                        pseudonym,
-                        dead=death_manager.is_dead(assassin_model),
-                        incompetent=competency_manager.is_inco_at(assassin_model, e.datetime),
-                        is_police=assassin_model.is_police,
-                        is_wanted=wanted_manager.is_player_wanted(assassin_model.identifier, time=e.datetime)
-                    )
-
-                    candidate_pseudonyms.append((assassin_model, pseudonym, color))
-
-            for (assassin_model, pseudonym, color) in candidate_pseudonyms:
-                headline = substitute_pseudonyms(headline, pseudonym, assassin_model, color, e.datetime)
-                for (k, r) in reports.items():
-                    reports[k] = substitute_pseudonyms(r, pseudonym, assassin_model, color, e.datetime)
+            headline, reports = render_headline_and_reports(
+                e,
+                death_manager=death_manager,
+                competency_manager=competency_manager,
+                wanted_manager=wanted_manager
+            )
 
             report_list = []
             for ((assassin, pseudonym_index), r) in reports.items():
@@ -289,13 +330,8 @@ class PageGeneratorPlugin(AbstractPlugin):
                 report_list.append(
                     REPORT_TEMPLATE.format(PSEUDONYM=painted_pseudonym, REPORT_COLOR=color, REPORT=r)
                 )
-            days_since_start = (e.datetime.date() - start_date).days
 
-            week = days_since_start // 7 + 1
-            day = days_since_start % 7
-            if week < 0:
-                week = 0
-                day = days_since_start + 7
+            days_since_start, week, day = date_to_weeks_and_days(start_date, e.datetime.date())
             events_for_chapter.setdefault(week, {})
 
             report_text = "".join(report_list)
@@ -309,8 +345,7 @@ class PageGeneratorPlugin(AbstractPlugin):
             events_for_chapter[week].setdefault(day, []).append(event_text)
 
             headline_text = HEAD_HEADLINE_TEMPLATE.format(
-                NUMBER=f"{week:02}",
-                ID=e._Event__secret_id,
+                URL=event_url(e),
                 TIME=time_str,
                 HEADLINE=headline
             )
--- a/AU2/plugins/custom_plugins/ScoringPlugin.py
+++ b/AU2/plugins/custom_plugins/ScoringPlugin.py
@@ -1,37 +1,170 @@
 import os
-from typing import List, Optional, Tuple, Any, Dict
+import pathlib
+import datetime
+from typing import List, Optional, Tuple, Any, Dict, Iterable
 
 from AU2 import ROOT_DIR
 from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
 from AU2.database.EventsDatabase import EVENTS_DATABASE
 from AU2.database.GenericStateDatabase import GENERIC_STATE_DATABASE
+from AU2.database.model.Event import Event
 from AU2.html_components.HTMLComponent import HTMLComponent
 from AU2.html_components.SimpleComponents.Label import Label
 from AU2.html_components.SimpleComponents.OptionalDatetimeEntry import OptionalDatetimeEntry
 from AU2.html_components.SimpleComponents.DefaultNamedSmallTextbox import DefaultNamedSmallTextbox
 from AU2.html_components.SimpleComponents.FloatEntry import FloatEntry
 from AU2.html_components.SimpleComponents.HiddenTextbox import HiddenTextbox
+from AU2.html_components.SimpleComponents.SelectorList import SelectorList
+from AU2.html_components.SimpleComponents.Checkbox import Checkbox
+from AU2.html_components.SimpleComponents.InputWithDropDown import InputWithDropDown
 from AU2.plugins.AbstractPlugin import AbstractPlugin, Export, ConfigExport
 from AU2.plugins.CorePlugin import registered_plugin
+from AU2.plugins.custom_plugins.PageGeneratorPlugin import get_color, render_headline_and_reports, event_url
 from AU2.plugins.constants import WEBPAGE_WRITE_LOCATION
 from AU2.plugins.util.ScoreManager import ScoreManager
-from AU2.plugins.util.date_utils import get_now_dt, timestamp_to_dt, dt_to_timestamp
+from AU2.plugins.util.CompetencyManager import CompetencyManager
+from AU2.plugins.util.WantedManager import WantedManager
+from AU2.plugins.util.DeathManager import DeathManager
+from AU2.plugins.util.date_utils import get_now_dt, timestamp_to_dt, dt_to_timestamp, DATETIME_FORMAT
+from AU2.plugins.util.game import get_game_start
 
-TABLE_TEMPLATE = """
+OPENSEASON_TABLE_TEMPLATE = """
 <table xmlns="" class="playerlist">
   <tr><th>Real Name</th><th>Address</th><th>College</th><th>Room Water Weapons Status</th><th>Notes</th><th>Points</th></tr>
   {ROWS}
 </table>
 """
 
-ROW_TEMPLATE = """
+OPENSEASON_ROW_TEMPLATE = """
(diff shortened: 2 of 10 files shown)


AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -156,6 +179,55 @@ def substitute_pseudonyms(string: str, main_pseudonym: str, assassin: Assassin,
...
+                                competency_manager: CompetencyManager,
+                                wanted_manager: WantedManager) -> Tuple[str, List[str]]:
+    headline = e.headline
+
+    reports = {(playerID, pseudonymID): soft_escape(report) for (playerID, pseudonymID, report) in
+               e.reports}
+
+    candidate_pseudonyms = []
+
+    texts_to_search = [r for r in reports.values()] + [headline]
```
jsyiek, line 193: doesn't need to be change as it's so short, but my C++ brain wants to point out `itertools.chain` is better in situations like this for joining list literals to avoid extra allocations

_(...this is my code that's been moved up, isn't it? oops...)_

jsyiek (commented):
a quick glance, will look properly another day

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -82,9 +82,16 @@
     "#ff0033", "#cc3333", "#ff3300"
 ]
 
+
+def event_url(e: Event, week: Optional[int] = None) -> str:
```
jsyiek, line 86: I'd put a brief doc string here. Why do we allow users to override week?

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -103,6 +110,22 @@
 FORMAT_SPECIFIER_REGEX = r"\[[D,P,N]([0-9]+)(?:_([0-9]+))?\]"
 
 
+def date_to_weeks_and_days(start: datetime.date, date: datetime.date):
+    """
+    Converts a datetime.date into a tuple of:
+        days since `start`
+        week number
+        day of the week
+    """
```
jsyiek, line 119:
```suggestion
    Converts param `date` into a tuple of:
        days since param `start`
        week number
        day of the week
    """
```

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -103,6 +110,22 @@
 FORMAT_SPECIFIER_REGEX = r"\[[D,P,N]([0-9]+)(?:_([0-9]+))?\]"
 
 
+def date_to_weeks_and_days(start: datetime.date, date: datetime.date):
```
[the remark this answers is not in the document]
jsyiek: I'm of the opinion the return type should be a `NamedTuple`, actually

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
           errors="ignore") as F:
     OPENSEASON_PAGE_TEMPLATE = F.read()
 
+# TODO: split up go generate table and row templates based on options
```
jsyiek, line 45: I don't understand this comment

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+
+
+def stats_table_template(columns: Iterable[str]) -> str:
+    """Generate a table template for the stats page from the given columns"""
+    return """
+<table xmlns="" class="playerlist">
+<tr>""" + "".join(f"<th>{header}</th>" for header in columns) + """</tr>
+  {ROWS}
+</table
+"""
```
jsyiek, line 72: A few things here:
1) Your final table tag is unclosed.
2) HTML doesn't care about the newlines here
3) The indentation is a little hard to follow

How about something like this?

```suggestion
    return (
        '<table xmlns="" class="playerlist"><tr>'
        + ''.join(f'<th>{header}</th>' for header in columns)
        + '</tr> {ROWS} </table>'
   )
```

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+
+
+STATS_ORDERING_KEYS = {
+    "By Death (pre-2022 style)": lambda score_manager, a: (-score_manager.get_rating(a), -score_manager.get_score(a)),
+    "By Kills (post-2022 style)": lambda score_manager, a: (-score_manager.get_kills(a), -score_manager.get_conkers(a)),
```
jsyiek, line 77: Is it called post/pre-2022 to urge umps to pick the latter? You could call it "London style" and "AU1 style" otherwise... just whether or not you want to give Henry some extra street cred :-)

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+}
+
+STATS_PAGE_TEMPLATE: str
+with open(os.path.join(ROOT_DIR, "plugins", "custom_plugins", "html_templates", "stats.html"), "r", encoding="utf-8",
+          errors="ignore") as F:
+    STATS_PAGE_TEMPLATE = F.read()
```
jsyiek, line 83: I've been playing with `pathlib.Path` and I prefer its readability to `os.path` (my previous go-to). What do you think about the following (assuming `ROOT_DIR` is changed to a `pathlib.Path`)? 

iff we are in agreement this style is nicer, maybe you could update `ROOT_DIR`, and we can do a proper refactor from `os.path` to `pathlib.Path`

```
STATS_PAGE_TEMPLATE: str
STATS_PAGE_PATH: pathlib.Path = ROOT_DIR / "plugins" / "custom_plugins" / "html_templates" / "stats.html"
with **open(STATS_PAGE_PATH, "r", encoding="utf-8", errors="ignore") as F:
    STATS_PAGE_TEMPLATE = F.read()
```

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+
+KILLTREE_PATH = "killtree.html"
```
jsyiek, line 85: not stored in the HTML folder?

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+<h2>Kill tree</h2>
+<script type="text/javascript">
+  function iframeLoaded() {
+      var iFrameID = document.getElementById('idIframe');
+      if(iFrameID) {
+            iFrameID.height = iFrameID.contentWindow.document.body.scrollHeight + "px";
+      }   
+  }
+</script>   
+""" + f"""
```
jsyiek, line 96: but why not just make this one single f-string?

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -33,6 +42,66 @@
...
+<iframe src="{KILLTREE_PATH}" width="100%" style="border: none; aspect-ratio: 4 / 3" scrolling="no" id="idIframe" onload="iframeLoaded()"></iframe>
+"""
+KILLTREE_LINK = f"""
+<p>View a visualisation of the kill graph <a href={KILLTREE_PATH}>here</a>.</p>
+"""
+
+DATETIME_FORMAT = "%Y-%m-%d %H:%M"
```
jsyiek, line 103: this sounds like a library variable. why not `KILLTREE_DATETIME_FORMAT` or something like that?

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                                                title="Enter when Open Season ended (this is used to determine which "
+                                                      "players were duellists)",
+                                                default=end))
+        return components
+
+    def answer_stats_page(self, htmlResponse) -> List[HTMLComponent]:
+        components = []
+        # this is silly but needed to fix a bug where columns end up in the wrong order
+        columns: List[str] = [c for c in STATS_COLUMN_TEMPLATES.keys()
+                              if c in htmlResponse[self.html_ids["Stats Columns"]]]
```
jsyiek, line 244: `.keys()` is redundant for dicts - that's the default when iterating over them

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+        stats_order = htmlResponse[self.html_ids["Stats Order"]]
+        self.gsdb_set("Stats Order", stats_order)
+        # use a score manager to count kills, conkers, and attempts
+        # don't need to set the formula because we aren't going to fetch scores
+        full_players = ASSASSINS_DATABASE.get_filtered(include=lambda a: not a.is_police,
+                                                       include_hidden=lambda a: not a.is_police)
+        formula = self.gsdb_get("Formula")
+        score_manager = ScoreManager({a.identifier for a in full_players}, formula=formula, game_end=openseason_end)
+        events = sorted(EVENTS_DATABASE.events.values(),
+                        key=lambda e: e.datetime)
```
jsyiek, line 259:
```suggestion
        events = sorted(EVENTS_DATABASE.events.values(), key=lambda e: e.datetime)
```

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+        for e in events:
+            score_manager.add_event(e)
+        rows = []
+        # "rating" here is a way of ordering players who died before the duel by time of death,
```
jsyiek, line 263: "rating" but rating isn't a variable. do you mean "rank"?

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                NAME=p.real_name,
+                PSEUDONYMS=p.all_pseudonyms(),
+                KILLS=score_manager.get_kills(p),
+                CONKERS=score_manager.get_conkers(p),
+                ATTEMPTS=score_manager.get_attempts(p),
+                DEATHS='<br />'.join(deaths) if deaths else "&mdash;",
+                RANK=(rank+1
+                      if rank == 0 or
+                      STATS_ORDERING_KEYS[stats_order](score_manager, full_players[rank-1])
+                        != STATS_ORDERING_KEYS[stats_order](score_manager, p)
```
jsyiek, line 281: this needs for loop, and particular these two lines, need comments explaining what's going on

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                                     ),
+                                     title=f"[{e.datetime.strftime(DATETIME_FORMAT)}] {plaintext_headline}"
+                        )
+                with open(os.path.join(WEBPAGE_WRITE_LOCATION, KILLTREE_PATH), "w+", encoding="utf-8") as F:
+                    F.write(net.generate_html())
+                killtree_embed = KILLTREE_EMBED
+                killtree_link = KILLTREE_LINK
+                components.append(Label("[SCORING] Generated killtree page."))
+            except ModuleNotFoundError as err:
+                components.append(Label("[WARNING] [SCORING] {err} -- skipping kill tree visualisation."))
```
jsyiek, line 355: missing f-specifier

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                SCORE=score_manager.get_score(p)
+            ))
+        table_str = stats_table_template(columns).format(ROWS="".join(rows))
+
+        # kill tree visualiser
+        killtree_embed = ""
+        killtree_link = ""
+        if generate_killtree:
+            # skip if we don't have pyvis or lxml installed
+            try:
```
jsyiek, line 292: a few things:
1) wrap only the minimal amount of code in the `try` - just the two module imports. 
2) not keen on local imports if we can avoid it

how about we move this logic to top-level?

```
PYVIS_ENABLED: bool = False
try: 
     from pyvis.network import Network
     import lxml.html
     PYVIS_ENABLED = True
except ModuleNotFoundError:
    pass
```

then, you can skip the below logic in the event you don't have it enabled

```
# instead of `if generate_killtree`
if generate_killtree and PYVIS_ENABLED:
# or better yet, find a way to do negative escapes - i.e., `if not PYVIS_ENABLED: continue`
# but I recognize this isn't in a loop so maybe this section needs to be its own function(?)
```

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                                title=victim_searchable,
+                                value=1+score_manager.get_conkers(victim_model)
+                            )
+                            added_nodes.add(victim)
+                        headline = render_headline_and_reports(
+                            e,
+                            death_manager=DeathManager(),
+                            competency_manager=competency_manager,
+                            wanted_manager=wanted_manager
+                        )[0]
```
jsyiek, line 337: alternatively, `*_` if you get more than one item out of this function. make it obvious at the front you're getting more than one thing out of this function and throwing all but one thing away

```suggestion
                        headline, _ = render_headline_and_reports(
                            e,
                            death_manager=DeathManager(),
                            competency_manager=competency_manager,
                            wanted_manager=wanted_manager
                        )
```

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -131,6 +214,160 @@ def answer_set_bonuses(self, html_response: Dict[str, Any]) -> List[HTMLComponen
...
+                        plaintext_headline = lxml.html.fromstring("<html>"+headline+"</html>").text_content()
```
jsyiek, line 338:
```suggestion
                        plaintext_headline = lxml.html.fromstring(f"<html>{headline}</html>").text_content()
```

AU2/plugins/util/ScoreManager.py
```diff
@@ -64,8 +70,38 @@ def _score(self, identifier: str) -> float:
...
+    def get_kills(self, a: Assassin) -> int:
+        return self._kills(a.identifier)
+
+    def get_attempts(self, a: Assassin) -> int:
+        return self._attempts(a.identifier)
+
+    def get_death_events(self, a: Assassin) -> List:
+        return self.deaths[a.identifier]
+
+    def get_rating(self, a: Assassin, game_end: Optional[datetime.datetime] = None) -> float:
```
jsyiek, line 101: also, why do we have this indirection? why not just have `_rating` take in the assassin directly and use `a.identifier` instead of `identifier`? 

```suggestion
    def get_rating(self, a: Assassin) -> float:

```

AU2/plugins/util/ScoreManager.py
```diff
@@ -64,8 +70,38 @@ def _score(self, identifier: str) -> float:
         b = self._bonus(identifier)
         a = self._attempts(identifier)
         import math
-        score = eval(self.formula)
+        score = eval(self.formula) if self.formula else c
         return score
 
+    @functools.cache
+    def _rating(self, identifier: str) -> float:
```
jsyiek, line 77: this is where you need to describe what "rating" is with a doc string

jsyiek (changes requested):
first proper pass over, thanks for making the stats page!

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
thenewtonium: To be honest this is more of an artifact of taking this section out of `on_page_generate`; my rationale was that it would allow `on_page_generate` to continue to use its own values for week numbers when generating urls on the headlines page. But there's not really any good reason to do this I suppose, so I'll remove that.

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Oops
I wrote this before I'd refactored my original code to allow for configuring which columns appear in the stats table and forgot to delete afterwards.

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Looks good! I don't know how I missed the unclosed `table` tag...

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Well it was mostly to give point of reference of why these are the two options... but I suppose my naming would become confusing if umpires start using the AU1 ordering again XD
I like the "London style" and "AU1" style suggestions though, I think the extra street cred for Henry is well-earned :)

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Yep I agree that that does look nicer. Although I think I'll call the path variable `STATS_PAGE_TEMPLATE_PATH` because to me `STATS_PAGE_PATH` makes it sound like the location of the generated page rather than the template ¯\\\_(ツ)\_/¯

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: because the `{` and `}` in the JavaScript would mess with it. I know I can escape them but it seemed better to be able to have the JS exactly as it will appear in the generated page.

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Good point... `DATETIME_FORMAT` is defined 3 times now and is always the same, I'll move it into `plugins/util/date_utils.py` — I don't think we need to have a datetime format defined specifically for the killtree.

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Oops
This is another stray comment from before I made things configurable, that I forgot to delete (originally the sorting by rating was done explicitly here rather than through `STATS_ORDERING_KEYS`)

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Something like this?
```python3
                # RANK is simply the position that a player appears according to whichever metric we are using,
                # e.g. "kills" or "rating". 
                # But if players are tied according to this ordering metric, we want to make that clear.
                # We do this by replacing a player's rank with a " -- representing a "ditto" sign -- if they are listed
                # below a player they are tied with.
                RANK=(rank+1
                      if rank == 0 or
                      STATS_ORDERING_KEYS[stats_order](score_manager, full_players[rank-1])
                        != STATS_ORDERING_KEYS[stats_order](score_manager, p)
                      else '"'),
```

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: Sure, that seems reasonable

thenewtonium: Hmm, one issue with this is it does seem to make AU2 take quite a bit longer to start up. It may be better to do a local import after all (but still wrap less code in `try`)

AU2/plugins/custom_plugins/ScoringPlugin.py
thenewtonium: I'm not sure what you mean?

AU2/plugins/util/ScoreManager.py
thenewtonium: The purpose of the indirection is to be able to cache the value of `_rating`, which seemed reasonable at first because it was calling `_conkers` (which isn't cached) rather than `_score`
but yes the indirection is a bit silly as it is now.
I'll remove `_rating` and just put the code into `get_rating`. I would get rid of `_kills`, `_attempts` and `_bonus` but that's complicated by the fact they're used in tests.

AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -100,7 +100,71 @@ def stats_table_template(columns: Iterable[str]) -> str:
 <p>View a visualisation of the kill graph <a href={KILLTREE_PATH}>here</a>.</p>
 """
 
-DATETIME_FORMAT = "%Y-%m-%d %H:%M"
+NODE_SHAPE = "dot"
+def generate_killtree_visualiser(events: List[Event], score_manager: ScoreManager) -> str:
```
jsyiek, line 104: Not a fan of string returns. Let's try something else, give me a few min to write you some code

jsyiek: Been meaning to get the Try monad the codebase and now is a good time - can you add something like this into a common location, implementing the relevant methods in the subclasses?

```
import abc
from typing import Callable, Generic, TypeVar


ResultT = TypeVar("Result")
ErrorT = TypeVar("ErrorT")
TransformT = TypeVar("TransformT")


class Result(abc.ABC, Generic[ResultT, ErrorT]):
    def get_or_else(self, or_else: Callable[[ErrorT], ResultT]) -> ResultT:
        """
        If contains success, then returns held result, else invokes passed callable
        """
        ...

    def map(self, fn: Callable[[ResultT], TransformT]) -> Result[TransformT, ErrorT]:
        """
        Transforms the held value, if any, using the input callable
        """
        ...

    def map_failure(self, fn: Callable[[ErrorT], None]) -> None:
        """
        Applies `fn` if an error type is held
        """
        ...
    
    def or_throw(self, fn: Callable[[ErrorT], Exception]) -> None:
        """
        Throws exception generated by fn if does not contain a success type
        """
        
        
class Success(Result):
    ...

class Failure(Result):
    ...
```

jsyiek: Then, this function can return a `Result[None, str]`, and the caller can pass a lambda to `map_failure`. Your client code looks a lot nicer using it, and I can go around the codebase at a later date and clean up a bunch of places to use this

Maybe we actually want a `map_all(fn1, fn2)` which is equiv to `map(fn1).map_failure(fn2)` idk

jsyiek: If needed this can be another PR but the Try monad is very fun to implement so happy for you take it given you have a good use case for it

jsyiek (approved):
approved +/- Try monad question

jsyiek:
Squash commits before you merge, please

```
git rebase -i master
```
(then change each 'pick' aside from the first to `squash` or `fixup`)

Issue #74, opened 2024-11-18
Sort Exports (and ConfigExports) for ease of use


NerdyNinja11:
Currently, Exports and ConfigExports are sorted alphabetically in the ui.
Once you have a lot of plugins enabled, these can get quite cluttered, and it's annoying to press the arrow keys 10s of times to select the options you want.

Either there could be a hardcoded sorting for some of the most common options ( eg Events -> Add Event, Generate Pages and Plugin config -> Plugin-specific parameters should probably be near the top)

Or there could be a system for the umpires to manually set the priorities of key Exports, with some sensible defaults essentially implementing the first solution.

thenewtonium:
> Don't we have something to this effect?
>
> #35 

I think there's still merit in being able to reorder exports. It's not necessarily the case that the other exports aren't getting used, and it's cumbersome to have to un-hide them every time you want to use them, then hide them again.

PR #101, opened 2025-02-04
[74] Allow frequently-used Exports to be pinned for ease of use

thenewtonium:
**Problem:** As described by @NerdyNinja11 in #74, now that we have more plugins and exports it can get unwieldy navigating through the menu. We do already have a feature to hide certain menu items, but for functionality that is used occasionally it is cumbersome to have to un-suppress an export to use it and then re-suppress it afterwards; it would be much more useful to be able to keep most of the exports visible on the main menu but put the most frequently used ones at the top.

**Solution:** I have implemented a simple "pinning" system for exports: the ConfigExport `CorePlugin -> Pin menu options` can be used to select exports to be "pinned". Pinned exports are listed before non-pinned exports in the main menu. (Within each category, exports are ordered by display name as before).
I believe this should be sufficient to improve usability, but the way I have implemented this is compatible with finer control over the order of exports: each export has a numerical 'priority' assigned to it in the `arb_state` of `CorePlugin`, which determines the ordering — with the current 'pinning' interface this simply takes a value of `0` (unpinned) or `1` (pinned).

I have also improved the interface for suppressing exports, so that display names of exports are displayed rather than the internal identifiers. Exports of disabled plugins also no longer appear in this interface even if they are currenty suppressed, but suppressed exports of disabled plugins will still remain suppressed.

**Testing:** Manual

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -4,7 +4,7 @@ import datetime
 import os
 import random
 import tabulate
-from typing import List, Any, Dict, Optional
+from typing import List, Any, Dict, Optional, Tuple
 
 import inquirer
 
@@ -743,18 +743,18 @@ def main():
     while True:
         core_plugin: CorePlugin = PLUGINS["CorePlugin"]
         exports = core_plugin.get_all_exports()
-
+        exit_option: List[Tuple[str, Optional[Export]]] = [("Exit", None)]
         q = [inquirer.List(name="mode", message="Select mode",
-                           choices=["Exit"] + [(e.display_name, e) for e in exports])]
+                           choices=exit_option + [(e.display_name, e) for e in exports])]
         try:
-            a = inquirer_prompt_with_abort(q)["mode"]
+            exp: Optional[Export] = inquirer_prompt_with_abort(q)["mode"]
         except KeyboardInterrupt:
-            a = "Exit"
-        if a == "Exit":
+            exp = None
+        if exp is None:
             print("Have a good day!")
             exit()
 
-        exp: Export = a
+        exp: Export
 
         params = []
         kwargs = {}
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -745,7 +745,7 @@ def main():
         exports = core_plugin.get_all_exports()
 
         q = [inquirer.List(name="mode", message="Select mode",
-                           choices=["Exit"] + sorted([e.display_name for e in exports]))]
+                           choices=["Exit"] + [(e.display_name, e) for e in exports])]
         try:
             a = inquirer_prompt_with_abort(q)["mode"]
         except KeyboardInterrupt:
@@ -754,11 +754,7 @@ def main():
             print("Have a good day!")
             exit()
 
-        exp: Export = None
-        for e in exports:
-            if e.display_name == a:
-                exp = e
-                break
+        exp: Export = a
 
         params = []
         kwargs = {}
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import copy
 import datetime
-import os
+import itertools
 import random
 import tabulate
 from typing import List, Any, Dict, Optional, Tuple
@@ -743,14 +743,15 @@ def main():
     while True:
         core_plugin: CorePlugin = PLUGINS["CorePlugin"]
         exports = core_plugin.get_all_exports()
-        exit_option: List[Tuple[str, Optional[Export]]] = [("Exit", None)]
+        exit_val = None
+        exit_option: List[Tuple[str, Optional[Export]]] = [("Exit", exit_val)]
         q = [inquirer.List(name="mode", message="Select mode",
                            choices=exit_option + [(e.display_name, e) for e in exports])]
         try:
             exp: Optional[Export] = inquirer_prompt_with_abort(q)["mode"]
         except KeyboardInterrupt:
-            exp = None
-        if exp is None:
+            exp = exit_val
+        if exp == exit_val:
             print("Have a good day!")
             exit()
 
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import copy
 import datetime
-import itertools
 import random
 import tabulate
 from typing import List, Any, Dict, Optional, Tuple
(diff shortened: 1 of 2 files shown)


AU2/frontends/inquirer_cli.py
```diff
@@ -745,7 +745,7 @@ def main():
         exports = core_plugin.get_all_exports()
 
         q = [inquirer.List(name="mode", message="Select mode",
-                           choices=["Exit"] + sorted([e.display_name for e in exports]))]
+                           choices=["Exit"] + [(e.display_name, e) for e in exports])]
```
jsyiek, line 748: How does inquirer render these, and how does it return values for selection? Is the below type hint wrong?

thenewtonium: The first element of each pair is what is displayed to the user, the second element is the return value.
`inquirer.List` does support mixing strings and tuples despite its type hints not indicating this, but I suppose I ought to satisfy the type checker by replacing `"Exit"` with a tuple.

thenewtonium: Hmm, seems like it's actually the `+` operator, rather than `inquirer` which is causing the issues with the type checker. Changing `"Exit"` to a tuple doesn't help because then it doesn't like the fact that the second element of the tuple isn't an `Export`...

jsyiek: Use an optional[export]?

thenewtonium: I can't tell it what type to expect, is the issue. If I write, say
```python
choices=[("Exit", None)] + [(e.display_name, e) for e in exports])]
```
then `+` expects a `List[Tuple[str, None]]`, and there's nowhere to tell it otherwise I don't think?

thenewtonium: I guess setting a variable to `[("Exit", None)] ` and using this to explicitly state the type would solve this though.

AU2/frontends/inquirer_cli.py
```diff
@@ -743,22 +743,18 @@ def main():
     while True:
         core_plugin: CorePlugin = PLUGINS["CorePlugin"]
         exports = core_plugin.get_all_exports()
-
+        exit_option: List[Tuple[str, Optional[Export]]] = [("Exit", None)]
         q = [inquirer.List(name="mode", message="Select mode",
-                           choices=["Exit"] + sorted([e.display_name for e in exports]))]
+                           choices=exit_option + [(e.display_name, e) for e in exports])]
```
jsyiek, line 748: nit: `itertools.chain`?

AU2/plugins/CorePlugin.py
```diff
@@ -344,31 +351,72 @@ def get_all_exports(self) -> List[Export]:
...
         return exports
 
     def ask_suppress_exports(self):
-        export_identifiers = [export.identifier for export in self.get_all_exports()]
+        export_name_id_pairs = [(export.display_name, export.identifier)
+                                for export in self.get_all_exports(include_suppressed=True)
+                                if export.identifier not in self.IRREMOVABLE_EXPORTS]
         default_suppression = GENERIC_STATE_DATABASE.arb_state.get("CorePlugin", {}).get("suppressed_exports", [])
-        export_identifiers = sorted(list(set(export_identifiers + default_suppression)))
-        for exp in self.IRREMOVABLE_EXPORTS:
```
jsyiek, line 354: Why are we allowing users to remove irremovable exports, and then discarding their choice later? Don't offer it in the first place, surely?

AU2/plugins/CorePlugin.py
```diff
@@ -344,31 +351,72 @@ def get_all_exports(self) -> List[Export]:
...
-            if exp in suppressed_exports:
-                suppressed_exports.remove(exp)
-        GENERIC_STATE_DATABASE.arb_state.setdefault("CorePlugin", {})["suppressed_exports"] = suppressed_exports
+        new_suppressed_exports = set(htmlResponse[self.config_html_ids["Suppressed Exports"]])
+        current_suppression = set(GENERIC_STATE_DATABASE.arb_state.get("CorePlugin", {}).get("suppressed_exports", []))
+        # only update suppression for exports of loaded plugins
+        for exp_id in (exp.identifier for exp in self.get_all_exports(include_suppressed=True)):
+            if exp_id in new_suppressed_exports:
+                current_suppression.add(exp_id)
```
jsyiek, line 379:
```suggestion
        for exp in self.get_all_exports(include_suppressed=True):
            exp_id = exp.identifier
            if exp_id in new_suppressed_exports:
                current_suppression.add(exp_id)
```

AU2/plugins/CorePlugin.py
```diff
@@ -344,31 +351,72 @@ def get_all_exports(self) -> List[Export]:
...
+                defaults=default_pinned,
+                options=export_name_id_pairs
+            )
+        ]
+
+    def answer_reorder_exports(self, htmlResponse: dict[str, Any]) -> List[HTMLComponent]:
+        pinned_exports = set(htmlResponse[self.config_html_ids["Pinned Exports"]])
+        # leave any hidden/unloaded exports' priorities intact,
+        # and keep any priorities that are not 0 or 1 intact if they prioritise the correct exports
+        # (this is for forwards-compatibility)
```
jsyiek, line 412: nit: if you're thinking with forwards-compatibility, I'd do the sorting based on the integer priority rather than bool "is it > 0 or not"

jsyiek (changes requested):
specifically the comment on `IRREMOVABLE_EXPORTS`

AU2/plugins/CorePlugin.py
thenewtonium: They aren't offered as a choice in the first place -- the comprehension generating the choices has `if export.identifier not in self.IRREMOVABLE_EXPORTS`

Which, I suppose does make the code discarding those choices redundant now.

AU2/plugins/CorePlugin.py
thenewtonium: The actual sorting *is* done based on integer priority (see `CorePlugin.get_all_exports`). The stuff with the > 0 is so that existing priorities don't all get normalised to only 0 and 1 after using `CorePlugin -> Pin menu options`.

AU2/frontends/inquirer_cli.py
thenewtonium: `choices` doesn't seem to work with an iterator. I could wrap the whole thing in `list()` but that sees kind of silly.

AU2/plugins/CorePlugin.py
jsyiek: Ah, missed that

AU2/frontends/inquirer_cli.py
```diff
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
 import copy
 import datetime
-import os
+import itertools
```
jsyiek, line 4: nit: remove unused import?

jsyiek (approved):

PR #103, opened 2025-02-09
[PlayerImporterPlugin] Player importer from Google Sheets

thenewtonium:
**Problem:** The umpires this term have had issues with importing assassins from the signups spreadsheet. We eventually figured out that it was a file permissions issue. Since I do not know Windows well enough I was unable to advise how to fix this. In future this potential issue form part of the handover materials, but it's better for the umpires to not have to deal with this at all.

**Solution:** I have created an Import function which takes the URL of a (publicly accessible) Google Sheet and downloads the CSV of it, then imports from this.

I have also made the importer able to cope with spreadsheets containing blank rows above the headers, and added graceful handling of `PermissionError`s.

**Limitations:** Of course, this is reliant on Google's endpoints remaining reliable (which, given how many things they have shut down over the years, isn't a given!) So it will still be important that future umpires have access to information on how to deal with the file permissions issue.

**Testing:** Manual, using a Google sheet with real signup data.

### DIFF
--- a/AU2/plugins/custom_plugins/PlayerImporterPlugin.py
+++ b/AU2/plugins/custom_plugins/PlayerImporterPlugin.py
@@ -1,5 +1,7 @@
 import csv
+import urllib.request
 import os
+import re
 from typing import List, Optional
 
 from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
@@ -12,6 +14,7 @@ from AU2.plugins.AbstractPlugin import AbstractPlugin, Export
 from AU2.plugins.CorePlugin import registered_plugin
 from AU2.plugins.constants import WATER_STATUSES
 
+GSHEET_SHARE_LINK_PATTERN = re.compile(r"docs.google.com/spreadsheets/d/([^/]+)/")
 
 @registered_plugin
 class PlayerImporterPlugin(AbstractPlugin):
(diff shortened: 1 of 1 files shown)


AU2/plugins/custom_plugins/PlayerImporterPlugin.py
```diff
@@ -138,3 +144,61 @@ def answer_read_assassins_csv(self, htmlResponse) -> List[HTMLComponent]:
...
+            Label(
+                title="This player importer requires a specific spreadsheet format."
+            ),
+            Label(
+                title="Required fields:"
+            ),
+            Label(
+                title="     real_name,pseudonym,pronouns,email,college,address,water_status,notes,is_police"
+            ),
+            Label(title="('is_police' will check if the answer is 'yes' in lower case and set police accordingly)"),
```
jsyiek, line 173: tab indent to be aligned with above label?

surely we can just do `.lower()` instead of requiring it to be lowercase?

jsyiek (approved):
fix and you can merge

AU2/plugins/custom_plugins/PlayerImporterPlugin.py
thenewtonium: > tab indent to be aligned with above label?

are you asking whether the reason for the indent is to align with the above label, or are you suggesting that the indent should be larger to align with the above label?

> surely we can just do .lower() instead of requiring it to be lowercase?

indeed it seems we already this (I mostly just copy-pasted the `Label`s from the CSV importer), so that `Label` is in fact inaccurate.

jsyiek: Indentation here is about the text within the labels, which seems to have been removed in the latest commit.

PR #111, opened 2025-02-19
[92] Sanity check for when player not dead but [DX] or [NX] used.

thenewtonium:
This sanity check detects when pseudonym specifiers of the form `[DX]` or `[NX]` are used in the headline or reports of an event where the player with secret id `X` did not die in that event, and suggests replacing these with `[PX]` (for the same value of `X`).

### DIFF
--- a/AU2/plugins/sanity_checks/PlayerNotDead.py
+++ b/AU2/plugins/sanity_checks/PlayerNotDead.py
@@ -1,5 +1,5 @@
 import re
-from typing import List, Dict, Container
+from typing import List, Dict, Set
 
 from AU2.database.model import Event
 from AU2.database.AssassinsDatabase import ASSASSINS_DATABASE
@@ -17,7 +17,7 @@ class PlayerNotDead(SanityCheck):
 
     identifier = "Player_Not_Dead"
 
-    def _find_incorrect(self, string: str, dead_secret_ids: Container[str], fixes: Dict[str, str]):
+    def _find_incorrect(self, string: str, dead_secret_ids: Set[str], fixes: Dict[str, str]):
         # matches either [DX] or [NX]
         for match in DX_NX_PATTERN.findall(string):
             X = match[1]
@@ -25,7 +25,7 @@ class PlayerNotDead(SanityCheck):
                 fixes[match[0]] = X
 
     def _gather_incorrect(self, e: Event):
-        dead_secret_ids = {ASSASSINS_DATABASE.get(victim_id)._secret_id for _, victim_id in e.kills}
+        dead_secret_ids = set(ASSASSINS_DATABASE.get(victim_id)._secret_id for _, victim_id in e.kills)
         output = {}
         for (_, _, report) in e.reports:
             self._find_incorrect(report, dead_secret_ids, output)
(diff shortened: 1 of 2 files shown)


AU2/plugins/sanity_checks/PlayerNotDead.py
```diff
@@ -0,0 +1,58 @@
...
+        return output
+
+    def suggest_event_fixes(self, e: Event) -> List[Suggestion]:
+        suggestions = []
+        to_fix = self._gather_incorrect(e)
+        for original, secret_id in to_fix.items():
+            alist = ASSASSINS_DATABASE.get_filtered(include=lambda a: a._secret_id == secret_id,
+                                                    include_hidden=lambda a: a._secret_id == secret_id)
+            if len(alist) > 0:
+                assert len(alist) == 1
```
jsyiek, line 42: why not `if len(alist) == 1`? do we ever want to crash AU over a failed assert?

AU2/plugins/sanity_checks/PlayerNotDead.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _find_incorrect(self, string: str, dead_secret_ids: Container[str], fixes: Dict[str, str]):
+        # matches either [DX] or [NX]
+        for match in DX_NX_PATTERN.findall(string):
+            X = match[1]
+            if X not in dead_secret_ids:
+                fixes[match[0]] = X
+
+    def _gather_incorrect(self, e: Event):
+        dead_secret_ids = {ASSASSINS_DATABASE.get(victim_id)._secret_id for _, victim_id in e.kills}
```
jsyiek, line 28: nit nit

```suggestion
        dead_secret_ids = set(ASSASSINS_DATABASE.get(victim_id)._secret_id for _, victim_id in e.kills)
```

AU2/plugins/sanity_checks/PlayerNotDead.py
```diff
@@ -0,0 +1,58 @@
...
+
+class PlayerNotDead(SanityCheck):
+    """
+    Finds and detects cases where [DX] and [NX] appear,
+    despite the player being alive.
+    """
+
+    identifier = "Player_Not_Dead"
+
+    def _find_incorrect(self, string: str, dead_secret_ids: Container[str], fixes: Dict[str, str]):
```
jsyiek, line 20:
```suggestion
    def _find_incorrect(self, string: str, dead_secret_ids: Set[str], fixes: Dict[str, str]):
```

AU2/plugins/sanity_checks/PlayerNotDead.py
```diff
@@ -0,0 +1,58 @@
...
+            self._find_incorrect(report, dead_secret_ids, output)
+        self._find_incorrect(e.headline, dead_secret_ids, output)
+        return output
+
+    def suggest_event_fixes(self, e: Event) -> List[Suggestion]:
+        suggestions = []
+        to_fix = self._gather_incorrect(e)
+        for original, secret_id in to_fix.items():
+            alist = ASSASSINS_DATABASE.get_filtered(include=lambda a: a._secret_id == secret_id,
+                                                    include_hidden=lambda a: a._secret_id == secret_id)
```
jsyiek, line 40:
```suggestion
            assassin_match = lambda a: a._secret_id == secret_id
            alist = ASSASSINS_DATABASE.get_filtered(include=assassin_match, include_hidden=assassin_match)
```

AU2/plugins/sanity_checks/PlayerNotDead.py
thenewtonium: My thinking here was that with `len(alist) == 1` if we ever had `len(alist) > 1` then it would fail silently, and we would miss the fact that the database has somehow got screwed up.
*But* I suppose catching that here, in this sanity check, is not actually particularly useful seeing as it is unlikely to kick in before the duplication of ids becomes a problem.

PR #112, opened 2025-02-21
[108] Change Bounties system

thenewtonium:
Bounties are now implemented as a flag on `Event`s that mark them as being bounties; these events are then collated in `bounties.html` and rendered as they would be on the news pages.

### DIFF
--- a/AU2/plugins/custom_plugins/html_templates/bounties.html
+++ b/AU2/plugins/custom_plugins/html_templates/bounties.html
@@ -19,6 +19,10 @@
 
 	<!--Main Content-->   <h2>Bounties</h2>
 
+	<p xmlns="">
+    	This is the <b>List of Bountied players</b>. Being subject to a bounty does not make a player a licit target where they might otherwise not be. If you're a regular assassin and do not have them as a target, and this player is not Wanted or Incompetent, you will go Wanted for killing them! But maybe the bounty is worth it...?
+	</p>
+
 {CONTENT}
 
 <!--Footer-->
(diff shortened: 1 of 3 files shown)


AU2/plugins/custom_plugins/html_templates/bounties.html
```diff
@@ -19,13 +19,7 @@
 
 	<!--Main Content-->   <h2>Bounties</h2>
 
-
-<p xmlns="">
-    This is the <b>List of Bountied players</b>. Being on this list does not make a player a licit target where they might otherwise not. If you're a regular assassin and do not have them as a target, and this player is not Wanted or Incompetent, you will go Wanted for killing them! But maybe the bounty is worth it...?
```
jsyiek, line 24: This disclaimer should probably stay

jsyiek (commented):
lgtm, please put back the rules disclaimer somewhere (you can make it more concise if you'd like)

AU2/plugins/custom_plugins/html_templates/bounties.html
thenewtonium: Since the format and potential scope of the list of bounties has changed, how about changing the second sentence to

> Being subject to a bounty does does not make a player a licit target where they would otherwise not be. If you're a regular assassin and do not have them as a target, and this player is not Wanted or Incompetent, you will go Wanted for killing them! But maybe the bounty is worth it...?

Or maybe

> Being a target of a bounty [...]

?

jsyiek (approved):

PR #117, opened 2025-03-05
[EF] Support the same assassin reporting under different pseudonyms

jsyiek:
EBF is common for emergency bug fix

But what about emergency features??? Emergency feature??? I've gone with EF. Not sure.

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import copy
 import datetime
+import itertools
 import random
 import tabulate
 from typing import List, Any, Dict, Optional, Tuple
@@ -208,11 +209,15 @@ def render(html_component, dependency_context={}):
         assassins_mapping = dependency_context[dependent]
         if not assassins_mapping:
             return {html_component.identifier: [], "skip": True}
+
+        assassins_in_event = [(aId, *rest) for (aId, *rest) in html_component.assassins if aId in assassins_mapping]
+        options = list(itertools.chain.from_iterable([
+            [(aId, i, p) for (i, p) in enumerate(pseudonyms)] for (aId, pseudonyms) in assassins_in_event]))
         q = [inquirer.Checkbox(
             name="q",
             message="Reports (select players with reports)",
-            choices=list(assassins_mapping.keys()),
-            default=list(a[0] for a in html_component.default)  # default: List[Tuple[str, int, str]]
+            choices=options,
+            default=html_component.default  # default: List[Tuple[str, int, str]]
         )]
         reporters = inquirer_prompt_with_abort(q)["q"]
         results = []
@@ -231,14 +236,14 @@ def render(html_component, dependency_context={}):
                 assassin_model = ASSASSINS_DATABASE.get(a)
                 print(f"    ({assassin_model._secret_id}) {assassin_model.real_name}")
         for r in reporters:
-            key = (r, assassins_mapping[r])
+            key = r[:2]
             q = [inquirer.Editor(
                 name="report",
-                message=f"Report: {escape_format_braces(r)}",
+                message=f"Report: {escape_format_braces(str(r))}",
                 default=escape_format_braces(default_mapping.get(key, ''))
             )]
             report = inquirer_prompt_with_abort(q)["report"]
-            results.append((r, assassins_mapping[r], report))
+            results.append((r[0], r[1], report))
         return {html_component.identifier: results}
 
     # dependent component
--- a/AU2/plugins/custom_plugins/SRCFPlugin.py
+++ b/AU2/plugins/custom_plugins/SRCFPlugin.py
@@ -1,5 +1,6 @@
 import contextlib
 import datetime
+import keyring
 import os
 import re
 import time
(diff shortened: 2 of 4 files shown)


AU2/frontends/inquirer_cli.py
```diff
@@ -208,11 +209,23 @@ def render(html_component, dependency_context={}):
...
+        options = list(itertools.chain.from_iterable([
+            [(aId, i, p) for (i, p) in enumerate(pseudonyms)] for (aId, pseudonyms) in assassins_in_event]))
+
+        default_options = []
+        for (aId, idx, _) in html_component.default:
+            potential_matches = [(other_aId, p) for (other_aId, p) in assassins_in_event if other_aId == aId]
+            if potential_matches:
+                (_, p) = potential_matches[0]
+                idx_to_use = min(len(p), idx)
+                default_options.append((aId, idx_to_use, p[idx_to_use]))
```
thenewtonium, line 223: I don't really understand what the point of this bit is? The pseudonym list never gets shorter even if pseudonyms are 'deleted', and even if that somehow does happen the component doesn't transfer the report attributed to the now non-existent pseudonym to the 'fallback' pseudonym, so I don't really see the point of including the fallback in the default.

AU2/frontends/inquirer_cli.py
```diff
@@ -208,11 +209,23 @@ def render(html_component, dependency_context={}):
...
         q = [inquirer.Checkbox(
             name="q",
             message="Reports (select players with reports)",
-            choices=list(assassins_mapping.keys()),
-            default=list(a[0] for a in html_component.default)  # default: List[Tuple[str, int, str]]
+            choices=options,
+            default=default_options
```
thenewtonium, line 228: Might be nice to have a "cleaner" display for the options? Maybe something like
```suggestion
            choices=[(f"{t[0]} as {t[2]}", t) for t in options],
            default=[(f"{t[0]} as {t[2]}", t) for t in default_options]
```

AU2/frontends/inquirer_cli.py
```diff
@@ -222,23 +235,22 @@ def render(html_component, dependency_context={}):
...
             for a in assassins_mapping:
                 assassin_model = ASSASSINS_DATABASE.get(a)
                 print(f"    ({assassin_model._secret_id}) {assassin_model.real_name}")
         for r in reporters:
-            key = (r, assassins_mapping[r])
+            key = r[:2]
             q = [inquirer.Editor(
                 name="report",
-                message=f"Report: {escape_format_braces(r)}",
+                message=f"Report: {escape_format_braces(str(r))}",
```
thenewtonium, line 249: Ditto for here
```suggestion
                message=f"Report: {escape_format_braces(f'{r[0]} as {r[2]}')}",
```

AU2/frontends/inquirer_cli.py
```diff
@@ -169,7 +170,7 @@ def render(html_component, dependency_context={}):
 
     # dependent component
     elif isinstance(html_component, AssassinPseudonymPair):
-        assassins = [a[0] for a in html_component.assassins]
+        assassins = list(set([a[0] for a in html_component.assassins] + [a for a in html_component.default.keys()]))
```
thenewtonium, line 173: This shouldn't actually be an issue, since the `assassins` passed to `AssassinPseudonymPair` by `CorePlugin`'s `on_event_request_update` do include any hidden assassins that are already in the event. (This is precisely the reason why the `include_hidden` argument of `AssassinsDatabase.get_ident_pseudonym_pairs` is a `Callable` and not a `bool`). But an extra layer of protection doesn't hurt...

AU2/frontends/inquirer_cli.py
```diff
@@ -208,11 +209,23 @@ def render(html_component, dependency_context={}):
         assassins_mapping = dependency_context[dependent]
         if not assassins_mapping:
             return {html_component.identifier: [], "skip": True}
+
+        assassins_in_event = [(aId, *rest) for (aId, *rest) in html_component.assassins if aId in assassins_mapping]
+        options = list(itertools.chain.from_iterable([
+            [(aId, i, p) for (i, p) in enumerate(pseudonyms)] for (aId, pseudonyms) in assassins_in_event]))
```
thenewtonium, line 215: It is slightly cursed that this includes deleted pseudonyms. Imo it would be better to filter them out here, except for any that appear in `default_options`.

E.g.
```suggestion
        options = list(itertools.chain.from_iterable([
            [(aId, i, p) for (i, p) in enumerate(pseudonyms) if p or (aId, i, p) in default_options]
            for (aId, pseudonyms) in assassins_in_event]))
```
after moving the code filling out `default_options` above this.

AU2/plugins/custom_plugins/SRCFPlugin.py
```diff
@@ -1,5 +1,6 @@
 import contextlib
 import datetime
+import keyring
```
thenewtonium, line 3: (Accidental commit?)

thenewtonium (approved):
Lgtm in terms of being functional, but some comments

jsyiek:
I'm not desperate to merge this yet, for the reasons outlined in #127. Let's park it.

PR #120, opened 2025-03-07
[ECF] Restore old `BountyPlugin` -- renaming rewritten bounty plugin to `BountyNewsPlugin`

thenewtonium:
Where "ECF" stands for "Emergency Compatibility Fix". It is only urgent in the sense that this needs to be merged before a new release is produced.

**Problem:** When I rewrote `BountyPlugin` in #112 I removed compatibility with the old style of bounties, as that feature had been unused. However, the umpires of the current game have now set a bounty using a version of `AU2` that does not have the rewrite.

**Solution:** This PR restores the old `BountyPlugin` and renames the rewritten bounty plugin to `BountyNewsPlugin` (I chose this name because the format of the bounties page in the rewrite is the same as the news pages). The pages generated by the two plugins are also now different — the pre-rewrite `BountyPlugin` keeps `bounties.html`, and `BountyNewsPlugin` generates `bounty-news.html` instead.

**Testing:** Both plugins have already been tested so I just did a cursory sanity-check making sure both plugins generated their respective pages without any errors, which they did.


jsyiek:
Do we still need this anymore?

thenewtonium:
> Do we still need this anymore?

I suppose not. (Especially if we merge #124).

PR #121, opened 2025-03-07
[116] Prevent creation of non-Police assassins after first email

thenewtonium:
**Problem:** It is currently possible to create a non-Police assassin once a game has started, even though this messes up the targeting graph.

**Solution:** Initially I was going to use `get_game_start` for this, but I realised that this wouldn't work because it defaults to the current datetime so is not a reliable indicator that that game has actually started. So instead I use the `last_emailed_event` stored in the GenericStateDatabase by `TargetingPlugin` when emails are sent, to determine whether the game has started or not. This is actually precisely what we want, because the whole point here is stopping the targeting graph completely changing, which only matters once targets have been sent out!

**Testing:** Manual.

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -472,7 +472,7 @@ def render(html_component, dependency_context={}):
         return {}
 
     elif isinstance(html_component, Checkbox):
-        default = "Yes" if html_component.checked else "No"
+        default = html_component.checked and "Yes" or "No"
         choices = [default] if html_component.force_default else ["No", "Yes"]
         q = [
             inquirer.List(
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -472,12 +472,14 @@ def render(html_component, dependency_context={}):
         return {}
 
     elif isinstance(html_component, Checkbox):
+        default = "Yes" if html_component.checked else "No"
+        choices = [default] if html_component.force_default else ["No", "Yes"]
         q = [
             inquirer.List(
                 name="q",
                 message=escape_format_braces(html_component.title),
-                choices=["No", "Yes"],
-                default="Yes" if html_component.checked else "No"
+                choices=choices,
+                default=default
             )]
         a = inquirer_prompt_with_abort(q)
         return {html_component.identifier: a["q"] == "Yes"}
--- a/AU2/html_components/SimpleComponents/Checkbox.py
+++ b/AU2/html_components/SimpleComponents/Checkbox.py
@@ -6,11 +6,12 @@ from AU2.html_components import HTMLComponent
 class Checkbox(HTMLComponent):
     name: str = "Checkbox"
 
-    def __init__(self, identifier: str, title: str, checked: bool=False):
+    def __init__(self, identifier: str, title: str, checked: bool=False, force_default: bool=False):
         self.title = title
         self.identifier = identifier
         self.uniqueStr = self.get_unique_str()
         self.checked = checked
+        self.force_default = force_default
         super().__init__()
 
     def _representation(self) -> str:
--- a/AU2/html_components/SimpleComponents/Checkbox.py
+++ b/AU2/html_components/SimpleComponents/Checkbox.py
@@ -6,7 +6,8 @@ from AU2.html_components import HTMLComponent
 class Checkbox(HTMLComponent):
     name: str = "Checkbox"
 
-    def __init__(self, identifier: str, title: str, checked: bool=False, force_default: bool=False):
+    def __init__(self, identifier: str, title: str, checked: bool=False, *, force_default: bool=False):
+
         self.title = title
         self.identifier = identifier
         self.uniqueStr = self.get_unique_str()
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -27,7 +27,7 @@ from AU2.plugins.AbstractPlugin import AbstractPlugin, Export, ConfigExport, Hoo
 from AU2.plugins.AvailablePlugins import __PluginMap
 from AU2.plugins.constants import COLLEGES, WATER_STATUSES
 from AU2.plugins.sanity_checks import SANITY_CHECKS
-from AU2.plugins.util.game import get_game_start, set_game_start
+from AU2.plugins.util.game import get_game_start, set_game_start, get_now_dt
 
 
 AVAILABLE_PLUGINS = {}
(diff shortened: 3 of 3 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -195,7 +195,17 @@ def __init__(self):
...
 
     def on_assassin_request_create(self):
+        # use this to detect whether the game has started or not, since sending the first email is the point when
+        # targets are "locked in" and adding new non-police players becomes dangerous
+        last_emailed_event = int(
+            GENERIC_STATE_DATABASE.arb_state.get("TargetingPlugin", {}).get("last_emailed_event", -1)
+        )
+
         html = [
+            *(
```
jsyiek, line 205: I feel like a cleaner approach would be to have the Checkbox take a parameter to conditionally remove one of the options 

Eg, taking a forceChoiceAs=Optional[Bool]. If set to true, it removes "No", and vice versa. 

This way, the umpire can see they have only one option instead of having a menu item scroll past them as a label. 

What do you think?

jsyiek (changes requested):

AU2/plugins/CorePlugin.py
thenewtonium: Yeah that does seem like a nicer way to to it

thenewtonium: I think I'll do it as a `force_default` attribute (defaulting to `False`) instead and use the `checked` attribute to decide which option to use, though.

AU2/frontends/inquirer_cli.py
```diff
@@ -472,12 +472,14 @@ def render(html_component, dependency_context={}):
         return {}
 
     elif isinstance(html_component, Checkbox):
+        default = "Yes" if html_component.checked else "No"
```
jsyiek, line 475: nit - I think I prefer having "yes" and "no" near each other in this case, although not super fussed

```suggestion
        default = html_component.checked and "Yes" or "No"
```

if you've not seen this style before, it looks like `default` will be a `bool`, but it's actually a `str`

```
>>> True and "Yes" or "No"
'Yes'
>>> False and "Yes" or "No"
'No'
```

AU2/plugins/CorePlugin.py
```diff
@@ -27,7 +27,7 @@
 from AU2.plugins.AvailablePlugins import __PluginMap
 from AU2.plugins.constants import COLLEGES, WATER_STATUSES
 from AU2.plugins.sanity_checks import SANITY_CHECKS
-from AU2.plugins.util.game import get_game_start, set_game_start
+from AU2.plugins.util.game import get_game_start, set_game_start, get_now_dt
```
jsyiek, line 30: nit (alphabetization) 

```suggestion
from AU2.plugins.util.game import get_game_start, get_now_dt, set_game_start

```

AU2/plugins/CorePlugin.py
```diff
@@ -195,7 +195,16 @@ def __init__(self):
...
+        # use this to detect whether the game has started or not, since sending the first email is the point when
+        # targets are "locked in" and adding new non-police players becomes dangerous
+        last_emailed_event = int(
+            GENERIC_STATE_DATABASE.arb_state.get("TargetingPlugin", {}).get("last_emailed_event", -1)
+        )
+
         html = [
+            Checkbox(self.html_ids["Police"], "Police? (y/n)",
+                     checked=last_emailed_event != -1,
+                     force_default=last_emailed_event != -1),
```
jsyiek, line 207: nit - unless we're expecting < -1? comment applies elsewhere as well

```
                     checked=last_emailed_event >= 0,
                     force_default=last_emailed_event >= 0),
```

AU2/plugins/CorePlugin.py
```diff
@@ -204,17 +213,23 @@ def on_assassin_request_create(self):
             InputWithDropDown(self.html_ids["Water Status"], "Water Status", WATER_STATUSES),
             InputWithDropDown(self.html_ids["College"], "College", COLLEGES),
             LargeTextEntry(self.html_ids["Notes"], "Notes"),
-            Checkbox(self.html_ids["Police"], "Police? (y/n)")
```
jsyiek, line 207: Why did you move it to the top? It's not crucial info to have first up

AU2/plugins/CorePlugin.py
```diff
@@ -243,6 +258,8 @@ def on_assassin_update(self, assassin: Assassin, htmlResponse: Dict) -> List[HTM
         assassin.water_status = htmlResponse[self.html_ids["Water Status"]]
         assassin.college = htmlResponse[self.html_ids["College"]]
         assassin.notes = htmlResponse[self.html_ids["Notes"]]
+        if self.html_ids["Police"] in htmlResponse:
+            assassin.is_police = htmlResponse[self.html_ids["Police"]]
```
jsyiek, line 262: Force default still populates the field in the response

```suggestion
        assassin.is_police = htmlResponse[self.html_ids["Police"]]
```

AU2/html_components/SimpleComponents/Checkbox.py
```diff
@@ -6,11 +6,12 @@
 class Checkbox(HTMLComponent):
     name: str = "Checkbox"
 
-    def __init__(self, identifier: str, title: str, checked: bool=False):
+    def __init__(self, identifier: str, title: str, checked: bool=False, force_default: bool=False):
```
jsyiek, line 9: force this to be a keyword arg for clarity (there's an argument `checked` should have been as well, but that's beyond this PR's scope)

```suggestion
    def __init__(self, identifier: str, title: str, checked: bool=False, *, force_default: bool=False):

```

AU2/plugins/CorePlugin.py
thenewtonium: Ah this is a hangover from when there was no interaction with the component I think

AU2/plugins/CorePlugin.py
```diff
@@ -223,17 +228,24 @@ def on_assassin_request_create(self):
...
     def on_assassin_create(self, assassin: Assassin, htmlResponse) -> List[HTMLComponent]:
         return [Label("[CORE] Success!")]
 
     def on_assassin_request_update(self, assassin: Assassin):
+        # use this to detect whether the game has started or not, since sending the first email is the point when
+        # targets are "locked in" and changing player types becomes dangerous.
+        last_emailed_event = int(
+            GENERIC_STATE_DATABASE.arb_state.get("TargetingPlugin", {}).get("last_emailed_event", -1)
+        )
+        print(last_emailed_event)
```
jsyiek, line 246: Delete

jsyiek (approved):
Good to merge if you delete print statement

PR #124, opened 2025-03-13
[EF/108] [`BountyPlugin`] Allow more flexible bounty targets (in a backwards-compatible way)

thenewtonium:
**Problem:** Dido requested the ability to set bounties on multiple people. This is relevant to [the bounty set by Gianpaolo Ruju](http://assassins.soc.srcf.net/news05.html#299), for example. The bounty system introduced by #112 does allow this but is incompatible with the bounty system being used this game.

**Solution:** Extend the existing `BountyPlugin` to allow an arbitrary string to be set as a target of a bounty, rather than forcing the user to select a player. When using `Bounties -> Update bounty` on old bounties the target string defaults to the real name of the existing target, and `Generate Pages` detects old bounties and renders them correctly.

**Testing:** Manual.

~~**Remarks:**  This PR is branched from #120, so that should be merged before this.~~


AU2/plugins/custom_plugins/BountyNewsPlugin.py
```diff
@@ -0,0 +1,115 @@
...
+
+from AU2 import ROOT_DIR
+from AU2.database.EventsDatabase import EVENTS_DATABASE
+from AU2.database.model.Event import Event
+from AU2.html_components import HTMLComponent
+from AU2.html_components.SimpleComponents.Checkbox import Checkbox
+from AU2.html_components.SimpleComponents.Label import Label
+from AU2.plugins.AbstractPlugin import AbstractPlugin, Export
+from AU2.plugins.CorePlugin import registered_plugin
+from AU2.plugins.custom_plugins.PageGeneratorPlugin import render_event, DAY_TEMPLATE, weeks_and_days_to_str, date_to_weeks_and_days
```
jsyiek, line 12: nit

```suggestion
from AU2.plugins.custom_plugins.PageGeneratorPlugin import date_to_weeks_and_days, render_event, weeks_and_days_to_str, DAY_TEMPLATE

```

AU2/plugins/custom_plugins/BountyNewsPlugin.py
```diff
@@ -0,0 +1,115 @@
...
+        e.pluginState.setdefault(self.identifier, {})["bounty"] = is_bounty
+        return [Label("[BOUNTY NEWS] Success")]
+
+    def on_event_update(self, e: Event, htmlResponse) -> List[HTMLComponent]:
+        is_bounty = htmlResponse[self.html_ids["bounty_event"]]
+        e.pluginState.setdefault(self.identifier, {})["bounty"] = is_bounty
+        return [Label("[BOUNTY NEWS] Success")]
+
+    def on_page_generate(self, _) -> List[HTMLComponent]:
+        events = list(EVENTS_DATABASE.events.values())
```
jsyiek, line 65:
```suggestion
        events = list(filter(lambda e: not e.hidden, EVENTS_DATABASE.events.values()))

```

AU2/plugins/custom_plugins/BountyNewsPlugin.py
```diff
@@ -0,0 +1,115 @@
...
+        is_bounty = htmlResponse[self.html_ids["bounty_event"]]
+        e.pluginState.setdefault(self.identifier, {})["bounty"] = is_bounty
+        return [Label("[BOUNTY NEWS] Success")]
+
+    def on_event_update(self, e: Event, htmlResponse) -> List[HTMLComponent]:
+        is_bounty = htmlResponse[self.html_ids["bounty_event"]]
+        e.pluginState.setdefault(self.identifier, {})["bounty"] = is_bounty
+        return [Label("[BOUNTY NEWS] Success")]
+
+    def on_page_generate(self, _) -> List[HTMLComponent]:
```
jsyiek, line 64: I feel like what we want is to take the page generation from the page generator plugin and to put it into a general library location, rather than duplicating it here?

AU2/plugins/custom_plugins/BountyNewsPlugin.py
thenewtonium: This isn't correct since hidden events still need to be added to the `CompetencyManager`, `WantedManager` and `DeathManager`.

AU2/plugins/custom_plugins/BountyNewsPlugin.py
thenewtonium: Good shout.

AU2/plugins/custom_plugins/BountyPlugin.py
```diff
@@ -1,115 +1,212 @@
...
+@dataclass
+class Bounty:
+    identifier: str = ""
+
+    # for backwards-compatibility, but note that creating a new bounty and loading the save in versions of AU2 up to
+    # v1.4.0-pre3 will cause a crash when using `Generate Pages` with `BountyPlugin` enabled, because those versions
+    # can't deal with a `Bounty`'s `target_id` being blank
+    target_id: str = ""
+
+    target_str: str = ""
```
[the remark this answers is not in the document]
thenewtonium: The reason was to maintain compatibility with old bounties, but tbf now that the lent game has finished it's no longer necessary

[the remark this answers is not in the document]
thenewtonium: To elaborate: the old bounties stored identifiers which are then rendered as real names. But the point of this PR is that the umpires would enter in the literal string to be rendered. So the old bounties after the update would have rendered the internal ids rather than the players' real names

AU2/plugins/custom_plugins/BountyPlugin.py
```diff
@@ -1,115 +1,212 @@
...
                 BOUNTIES_PAGE_TEMPLATE.format(
-                    CONTENT=bounty_content,
-                    YEAR=str(get_now_dt().year)
+                    YEAR=get_now_dt().year,
+                    TABLE_OPTIONAL=table_str
                 )
             )
         return [Label("[BOUNTY] Success!")]
+
+    def _read_bounty(self, bounty_id: str):
```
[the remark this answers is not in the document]
thenewtonium: I think the reason is because of the use of `self.identifier`

AU2/plugins/custom_plugins/BountyPlugin.py
```diff
@@ -1,115 +1,212 @@
...
+        return Bounty.from_dict(
+            bounty_id,
+            GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {}).setdefault("bounties", {}).setdefault(
+                bounty_id, {})
+        )
+
+    def get_bounty_ids(self):
+        return list(GENERIC_STATE_DATABASE.arb_state.setdefault(self.identifier, {}).setdefault("bounties", {}).keys())
```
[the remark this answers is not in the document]
thenewtonium: I don't think so (most of this is your old code restored btw...)

AU2/plugins/custom_plugins/html_templates/bounties.html
```diff
@@ -19,11 +19,13 @@
...
 
-	<p xmlns="">
-    	This is the <b>List of Bountied players</b>. Being subject to a bounty does not make a player a licit target where they might otherwise not be. If you're a regular assassin and do not have them as a target, and this player is not Wanted or Incompetent, you will go Wanted for killing them! But maybe the bounty is worth it...?
-	</p>
 
-{CONTENT}
+<p xmlns="">
+    This is the <b>List of Bountied players</b>. Being on this list does not make a player a licit target where they might otherwise not. If you're a regular assassin and do not have them as a target, and this player is not Wanted or Incompetent, you will go Wanted for killing them! But maybe the bounty is worth it...?
+</p>
+
```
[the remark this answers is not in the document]
thenewtonium: ¯\\\_(ツ)\_/¯ this is how it was before
realised I missed this in my response commits oops

AU2/plugins/util/game.py
```diff
@@ -1,15 +1,29 @@
...
+def soft_escape(string: str) -> str:
+    """
+    Escapes html and adds <br /> to newlines only if not prefixed by HTML_REPORT_PREFIX
+    """
+
+    # umpires may regret allowing this
+    # supposing you are a clever player who has found this and the umpire does not know...
+    # please spare the umpire any headaches
+    # and remember that code injection without explicit consent is illegal (CMA sxn 2/3)
+    if not string.startswith(HTML_REPORT_PREFIX):
```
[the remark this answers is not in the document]
thenewtonium: It would definitely be unambiguously a good idea if we had good documentation.

As is... I don't know.

AU2/plugins/constants.py
```diff
@@ -75,7 +76,7 @@
     "Full water"
 ]
 
-WEBPAGE_WRITE_LOCATION = os.path.expanduser("~/pages")
+WEBPAGE_WRITE_LOCATION = pathlib.Path(os.path.expanduser("~/pages"))
```
jsyiek, line 79:
```suggestion
WEBPAGE_WRITE_LOCATION = pathlib.Path("~/pages").resolve()

```

AU2/plugins/constants.py
```diff
@@ -75,7 +76,7 @@
     "Full water"
 ]
 
-WEBPAGE_WRITE_LOCATION = os.path.expanduser("~/pages")
+WEBPAGE_WRITE_LOCATION = pathlib.Path(os.path.expanduser("~/pages"))
 
 if not os.path.exists(WEBPAGE_WRITE_LOCATION):
     os.makedirs(WEBPAGE_WRITE_LOCATION)
```
jsyiek, line 82: missing newline at eof

jsyiek (approved):

Issue #123, opened 2025-03-08
HTML pseudonyms broken on wanted page


NerdyNinja11:
They are escaped, so render as raw html. Works on the police page, however

PR #130, opened 2025-03-20
[123] Changed escaping of list of pseudonyms to `soft_escape` each pseudonym individually

thenewtonium:
**Problem:** As identified in #123, HTML pseudonyms do not render on the Wanted page but do on the Police page. This is because inconsistent escaping has been applied.

**Solution:** Rather than remove the escaping altogether, I have modified the `Assassin.all_pseudonyms` method to take as an argument an escaping function to be applied to *each pseudonym individually*. The individual plugins then pass `soft_escape` into this function; this means that pseudonyms both with and without the HTML prefix are rendered correctly on the Wanted and Police pages. (That is, pseudonyms without the prefix are escaped but those with are not).

**Testing:** Manual. I have tested pseudonyms containing HTML both with and without the HTML prefix.


AU2/database/model/Assassin.py
```diff
@@ -202,5 +202,5 @@ def pseudonyms_until(self, ts: Optional[dt.datetime] = None) -> Iterator[str]:
             if p and self.pseudonym_datetimes.get(i, ts) <= ts:
                 yield p
 
-    def all_pseudonyms(self) -> str:
-        return " AKA ".join(p for p in self.pseudonyms if p)
+    def all_pseudonyms(self, escape: Callable[[str], str] = lambda x: x) -> str:
+        return " AKA ".join(escape(p) for p in self.pseudonyms if p)
```
jsyiek, line 206: An escape function that doesn't escape by default doesn't seem right to me. How about renaming this to `fn` and clarifying (with an appropriate doc string) that this function is mapped over pseudonyms?

jsyiek: Also, I feel like `soft_escape` should be the default value here

jsyiek (commented):
fine except for some comments on `all_pseudonyms`

jsyiek (approved):

PR #131, opened 2025-03-20
[122] Fixed list of corpses on inco list page

thenewtonium:
**Problem:** Currently, the "list of corpses" on the inco list page has been disabled. This is because of a logic error: the current code shows a list of *incos who are currently dead* rather than *players who died whilest inco*.

**Solution:** `CompetencyManager` now also tracks a list of corpses of players who died while inco. This only works correctly if events are added to the manager in chronological order (as `CompetencyPlugin` does).

**Testing:** Manual.
I created events both where a player died while inco and where a player died while not inco, then generated the page inco page for after the latter's inco deadline had passed. Only the former player was included in the list of corpses.


jsyiek (commented):
can you add a unit test or two for this feature please? otherwise looks fine.

jsyiek (approved):

PR #147, opened 2025-06-19
May week 2025 plugin

thenewtonium:
Some notes on the implementation (mainly for @axellerationduetogravity):

- 'Crews' are called 'Teams' and 'Mythic Compasses' are called 'Multipliers' by default; you can rename them using `May Week -> Tweak Cosmetics` under `Plugin config -> Plugin-specific parameters`.
- Internally, multipliers (mythic compasses) are held by players rather than teams, but with multiplier team sharing enabled (`May Week -> Enable/disable multiplier team sharing` in plugin config -- disabled by default) the score will be calculated and pages generated as if the team has the multiplier. This may make transfering multipliers a little fiddly since you always need to ensure the multiplier (mythic compass) holder is included in the event even if they weren't actually involved.
- Events are always processed in the order they were created rather than based on the datetimes assigned to each event (this isn't ideal but ensures consistency)
- Team names are set in the plugin config `May Week -> Rename Teams`, but team membership is set in events so that it can change without messing up scoring  -- you'll want to create an event at the start of the game setting all the initial teams.
- Bonuses can be assigned by giving 'BS points' in an event. These don't affect the points gained from kills in that event (but do affect subsequent events)


thenewtonium:
Actually, we probably should create a may week news page that has it all on one page rather than having `Week 1'

PR #150, opened 2025-07-15
[127] Improve report editing interface

thenewtonium:
**Problem:** This PR is to address some of the issues with the report editing interface in #117, namely
- When editing events it is tedious to click through all the reports
- It is easy to accidentally select the wrong pseudonym and hence leak it
- It is easy to accidentally delete a report
- There is no way of controlling the ordering of reports

**Solution:** I have rewritten the `AssassinDependentReportEntry` component to be more like `EditablePseudonymList`. The user is presented with a list of reports in the event as well as `*NEW*` and can edit existing reports and add new ones. For each report the user selects an author and one of their pseudonyms, or `(Auto)` which will automatically assign the pseudonym that has been set as the author's in the event. 

**Testing:** Manual.


thenewtonium:
Actually, I'm not sure about having the `(Auto)` option dealt with at page-generation time (e.g. it is possible to remove a player from an event without removing their report, in which case trying to generate pages causes a crash)

AU2/frontends/inquirer_cli.py
```diff
@@ -207,43 +207,108 @@ def render(html_component, dependency_context={}):
...
+            Renders the report editor UI for `report`, returning the edited report,
+            or `None` if the report should be deleted.
+            """
+            q = [inquirer.List(
+                name="author",
+                message="Select the AUTHOR of the report",
+                choices=[("*DELETE REPORT*", None), *((snapshot(ASSASSINS_DATABASE.get(a)), a) for a in assassins_mapping.keys())],
+                default=old_report[0]
+            )]
+            ident = inquirer_prompt_with_abort(q)["author"]
```
jsyiek: clarity

```suggestion
            ident: Optional[str] = inquirer_prompt_with_abort(q)["author"]
```

AU2/frontends/inquirer_cli.py
```diff
@@ -207,43 +207,108 @@ def render(html_component, dependency_context={}):
...
+            pseudonym_options = ((p, i) for i, p in enumerate(assassin_pseudonyms[ident]) if p)
+            q = [inquirer.List(
+                name="pseudonym",
+                message="Select the PSEUDONYM to attribute the report to",
+                choices=[("(Auto)", None), *pseudonym_options],
+                default=old_report[1]
+            )]
+            pseudonym_id = inquirer_prompt_with_abort(q)["pseudonym"]
+            pseudonym_id = int(pseudonym_id) if pseudonym_id is not None else assassins_mapping[ident]
+            pseudonym = assassin_pseudonyms[ident][pseudonym_id] if pseudonym_id is not None else None
```
jsyiek: Unconditional list de-ref in front-end leaves me worried... maybe `pseudonym_id is not None and len(assassin_pseudonyms[ident]) > pseudonym_id`

AU2/plugins/custom_plugins/SRCFPlugin.py
```diff
@@ -1,5 +1,6 @@
 import contextlib
 import datetime
+import keyring
```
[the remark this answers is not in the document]
thenewtonium: oh I think that was inherited from when I branched from #117

jsyiek (approved):

Issue #46, opened 2024-10-29
Cross-referencing kills with targeting graph


NerdyNinja11:
In AU1, it was annoying to have to check whether a kill was licit (i.e. it was either the killer's target or targetter)
In the end, I sort of didn't bother checking, running on the assumption that players would tell you if it wasn't obviously licit.

It would be really nice if AU2 could warn you if a just-added kill was illicit (purely based on the targeting graph), so you wouldn't miss kills that should make a player wanted

jsyiek:
Ideally it should tell you before you have the option to make anyone wanted but after you add kills. 

Here's an idea for how it could work: Add a `class StrictlyBefore(identifier_of_other: str, component_to_place_before: HTMLComponent)` meta-component that can be wrapped around an `HTMLComponent`. The front-end should respect this request - will require some `inquirer_cli.py` modification - but it's ok if cycles are unhandled (best-effort guarantee with no crashing if not handling cycles). Then, the `TargetingPlugin` should emit a new `DependentConditionalLabel` whose contents are determined by the selected assassins in the event and prints out the targeting graph for those assassins (also requires `inquirer_cli.py` modification). Then finally, all three of the conditional label, the kills component, and the wanted component should be in a strict ordering using `StrictlyBefore`. 

Other implementations fine.

Good ticket if you want something chunkier as a first-time issue - will have impact for the umpires while also having good creativity allowed.

NerdyNinja11:
The conditional inquire_cli stuff might make sense to do alongside #45

thenewtonium:
Just had a thought: an implementation that might make sense (and would be easier than rewriting inquirer_cli) would be to have this as part of `AssassinDependentCrimeEntry`, with an API call for "victim information" that wouldn't just give targets but also whether a victim was wanted or inco, and have `AssassinDependentCrimeEntry` also depend on kills (like the team kills entry from may week)

Issue #153, opened 2025-10-28
Include `police` in human-readable identifiers of police players


thenewtonium:
So that umpires can tell at a glance whether a player is police or not.

PR #154, opened 2025-10-28
[46] Include targeting info of victims in AssassinDependentCrimeEntry

thenewtonium:
**Problem:** Checking licitness of kills is somewhat annoying for umpires, which is something we've wanted to address for a while (see #46).

**Solution:** This PR adds the ability for `AssassinDependentCrimeEntry` to render information useful for determining licitness of kills. A new API call `render_licitness_info` has been added which returns a dictionary mapping player ids to a list of components for `AssassinDependentCrimeEntry` to render when a given player is selected as being the victim of a kill in an event. (Note that `AssassinDependentCrimeEntry` now also depends on `AssassinDependentKillEntry` like the `KillDependentSelector` from the May Week plugin)

In particular, the `render_licitness_info` method `TargetingPlugin` returns a mapping giving two `Table`s for each potential victim: one of the assassins who have them as targets and one of the assassins targeting them.

Also, the `render_licitness_info` method of `CorePlugin` returns a mapping giving `Label`s telling the user when a victim is police.

**Remark:** I wanted to also display the wantedness / inco status of victims, but to do this correctly in `Event -> Update` would require the licitness info to depend on the datetime set for the event. (This wasn't an issue for targets since they are calculated based on the order in which events are added by the umpire rather than when they happened...).  #143 would make it easy to do this but it's quite a big PR and needs to be updated to account for the May Week stuff.

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -353,7 +353,7 @@ def render(html_component, dependency_context={}):
                 elif killer in html_component.targeting_graph.get(victim, []):
                     print(f"{killer} was a target of {victim}.")
                 else:
-                    print(f"{killer} did not have {victim} as a target nor targetor; this may be an illicit kill.")
+                    print(f"{killer} did not have {victim} as a target nor targeter; this may be an illicit kill.")
 
         dependent = html_component.pseudonym_list_identifier
         assert (dependent in dependency_context)


thenewtonium:
> Why is html_components/.DS_Store here?

Accidental commit (oops). (.DS_Store is a flie MacOS automatically creates to store how files should be arranged visually when viewing a directory in Finder)

> This seems pretty good, my one thought is that it might be better to say explicitly that, for example,
'PlayerB was one of PlayerA's targets' or 'PlayerB was not a target or targeter of PlayerA; this may be an illicit kill',
instead of simply printing out the six relevent names, which is quite a lot of text.
> 
> Not sure how easy/desirable this would be, though. What do you think?

I think your suggestion would be better.

It would require taking a different approach to how `AssassinDependentCrimeEntry` receives licitness information, but the reason I took the approach I did was that I wanted to include Wantedness and Inco status as well, which turned out to be problematic because they are based on processing events in order of datetime rather than secret id, and the backend has no way of knowing what the datetime of a new event will be. So I suppose we might as well have `AssassinDependentCrimeEntry` explicitly take the targeting graph as a parameter and determine the relevant information as you suggest, seeing as it won't be able to deal with the other licitness criteria anyway (other than being police which as you say would be redundant with #153) without a significant reworking of how `Export`s work (e.g. #143)...

AU2/frontends/inquirer_cli.py
```diff
@@ -345,12 +345,15 @@ def render(html_component, dependency_context={}):
...
-            for v in victims:
-                for comp in html_component.licitness_info.get(v, []):
-                    render(comp)
+            for (killer, victim) in kills:
+                if victim in html_component.targeting_graph.get(killer, []):
+                    print(f"{killer} had {victim} as a target.")
+                elif killer in html_component.targeting_graph.get(victim, []):
+                    print(f"{killer} was a target of {victim}.")
+                else:
+                    print(f"{killer} did not have {victim} as a target nor targetor; this may be an illicit kill.")
```
NerdyNinja11: spelling, target**e**r

NerdyNinja11 (approved):
lgtm, other than the tiny spelling mistake.

Am I correct in thinking that this will not work in the unlikely scenario that:
1. PlayerA illicitly kills PlayerB, but delays reporting
2. PlayerC is killed, and this is reported first
3. PlayerB, previously targeting PlayerC, now targets PlayerA
4. When the (chronologically) first kill event is added, it appears as though PlayerA's kill was licit.
I assume nothing can be done about this

Issue #152, opened 2025-10-28
Make auto-competency and attempt tracking the default setting


thenewtonium:

NerdyNinja11:
As I mentioned in the review of #156, there may still be need for attempt tracking without competency for the purpose of scoring in May Week games

PR #156, opened 2025-10-29
[152] Clearer auto-competency

thenewtonium:
**Problem:** In this game the umpires had some confusion over whether attempts should be tracked manually, hence in #152 I suggested making auto-competency and attempt tracking the default setting. Thinking about this I also noticed a couple of other areas where auto-competency was unintuitive:
- attempt tracking is toggled separately to auto-competency
- in "Auto" (rather than "Full Auto") competency mode the role of the "Extend competency?" component is unintuitive

**Solution:** In this PR:
- "Auto" is now the default competency mode (and if unset this will be stored in `GenericStateDatabase` so that reverting to an older version of AU2 doesn't implicitly revert to "Manual" competency)
- attempt tracking is now based on whether auto-competency is enabled rather than a separate config (but the attempt tracking config is still updated in the back-end for backwards-compatibility)
- the orders of the components to add attempts and to extend competency have been switched so that attempts are asked about first. Also, when auto-competency is enabled the competency extension component is now labelled "Manually extend competency?", but when auto-competency is disabled it is still labelled simply "Extend competency?".

**Testing:** Manual.

### DIFF
--- a/AU2/plugins/custom_plugins/CompetencyPlugin.py
+++ b/AU2/plugins/custom_plugins/CompetencyPlugin.py
@@ -11,6 +11,7 @@ from AU2.database.GenericStateDatabase import GENERIC_STATE_DATABASE
 from AU2.database.model import Event, Assassin
 from AU2.html_components import HTMLComponent
 from AU2.html_components.DependentComponents.AssassinDependentIntegerEntry import AssassinDependentIntegerEntry
+from AU2.html_components.SimpleComponents.Checkbox import Checkbox
 from AU2.html_components.SimpleComponents.DatetimeEntry import DatetimeEntry
 from AU2.html_components.MetaComponents.Dependency import Dependency
 from AU2.html_components.SimpleComponents.InputWithDropDown import InputWithDropDown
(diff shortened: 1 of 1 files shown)


NerdyNinja11 (commented):
The reason attempt tracking was a separate toggle to autocompetency was that they were both new features to AU2, not in AU1, so we needed playtesting before deciding to make them default behaviour. Attempt tracking also affects scoring and gigabolt, so could theoretically have been wanted without autocompetency enabled.
(Maybe for May Week scoring or something?)

I would say that sensible defaults and actually writing some end-user documentation is more important than reducing flexibility.

Either way, nothing in this change is particularly urgent, so if we were to merge it, I think it makes sense to scrap the backwards compatible code bloat and merge after this game is over.

thenewtonium:
> Attempt tracking also affects scoring and gigabolt, so could theoretically have been wanted without autocompetency enabled.

That makes sense I guess. Maybe we should move attempt tracking into `CorePlugin` given this?

For now I'll revert this PR back to a having a seperate attempt tracking config but have it default to on.

NerdyNinja11 (approved):
lgtm

I agree that attempt tracking makes sense to be in `CorePlugin` since they can be useful for scoring without using competency; and maybe even be always on - I'm not sure, what does @jsyiek think?

PR #159, opened 2025-11-03
Confirmation step in `Event -> Delete`

thenewtonium:
**Problem:** Currently `Event -> Delete` doesn't require confirmation before deleting an event. SInce (currently at least) deleting an event can have irreversible effects on the targeting graph, and it would very much be possible to accidentally select the wrong event in the current interface, it seems to me that for safety there should be some sort of confirmation before actually deleting.

**Solution:** This PR requires the user to confirm deletion of an event in the same manner as `Reset Database`, i.e. by copying a string of numbers displayed to them. In the process of implementing this I also removed the `on_event_delete` and `on_event_request_delete` API calls as they are unused and their usage would be complicated by the confirmation step. We can always add them back later if the need arises....

**Testing:** Manual.

### DIFF
--- a/AU2/html_components/DerivativeComponents/DigitsChallenge.py
+++ b/AU2/html_components/DerivativeComponents/DigitsChallenge.py
@@ -33,7 +33,7 @@ def DigitsChallenge(identifier_prefix: str, title: str) -> List[HTMLComponent]:
     ]
 
 
-def read_DigitsChallenge(identifier_prefix: str, html_response: dict) -> bool:
+def verify_DigitsChallenge(identifier_prefix: str, html_response: dict) -> bool:
     """
     Reads the response from a pair of components produced by DigitsChallenge.
 
@@ -44,5 +44,7 @@ def read_DigitsChallenge(identifier_prefix: str, html_response: dict) -> bool:
     Returns
         bool: whether or not the user completed the challenge successfully
     """
-    return html_response[identifier_prefix + "_DigitsChallenge_HiddenTextbox"] == html_response[identifier_prefix + "_DigitsChallenge_NamedSmallTextbox"]
+    target = html_response[identifier_prefix + "_DigitsChallenge_HiddenTextbox"]
+    user_response = html_response[identifier_prefix + "_DigitsChallenge_NamedSmallTextbox"]
+    return target == user_response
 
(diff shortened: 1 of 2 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -81,7 +79,8 @@ def __init__(self):
             "Police": self.identifier + "_police",
             "Hidden Assassins": self.identifier + "_hidden_assassins",
             "Nuke Database": self.identifier + "_nuke",
-            "Secret Number": self.identifier + "_secret_confirm"
+            "Secret Number": self.identifier + "_secret_confirm",
+            "Delete Event": self.identifier + "_delete_event"
```
jsyiek: End with comma to avoid clogging diff like you've been forced to do here

AU2/plugins/CorePlugin.py
```diff
@@ -641,19 +634,27 @@ def gather_events(self) -> List[Tuple[str, str]]:
...
+            HiddenTextbox(self.HTML_SECRET_ID, event_id),
+            HiddenTextbox(
+                identifier=self.html_ids["Secret Number"],
+                default=str(i)
+            ),
+            Label(f"You are about to delete the event [{event.datetime.strftime('%Y-%m-%d %H:%M %p')}] {event.headline}"),
+            NamedSmallTextbox(
+                identifier=self.html_ids["Delete Event"],
+                title=f"Type {i} to confirm event deletion"
+            )
```
jsyiek: comma

AU2/plugins/custom_plugins/UIConfigPlugin.py
```diff
@@ -20,7 +20,6 @@ class Call(Flag):
     NONE = 0
     EVENT_CREATE = 2 << 0
     EVENT_UPDATE = 2 << 1
-    EVENT_DELETE = 2 << 2
```
jsyiek: Rename to RESERVED_ONE rather than deleting

AU2/html_components/DerivativeComponents/DigitsChallenge.py
```diff
@@ -0,0 +1,48 @@
...
+            default=str(i)
+        ),
+        NamedSmallTextbox(
+            identifier=identifier_prefix + "_DigitsChallenge_NamedSmallTextbox",
+            title=title.format(digits=i)
+        ),
+    ]
+
+
+def read_DigitsChallenge(identifier_prefix: str, html_response: dict) -> bool:
```
jsyiek: should probably be `verify` rather than `read`

Issue #140, opened 2025-03-22
`Setup Game` Export


jsyiek:
It occurs to me that we have a large number of parameters & configurations that are needed to start the game.

Maybe the `CorePlugin` needs a 'Set up game' export to walk the user through how the tool works and to help them set the basic config parameters?

thenewtonium:
Might be useful to also have this run automatically when AU2 is run for the first time (or rather when no existing database is detected).

PR #163, opened 2025-11-05
[140] `Setup Game` export

thenewtonium:
**Problem:** As noted in #140 there a number of parameters in AU2 that need to be set at the start of a game, so it would be useful to a feature to walk the user through setting these up.

**Solution:** This PR adds a `Setup Game` export and associated API calls `on_request_setup_game` and `on_setup_game`. The user is first presented with a list of game types (currently only "Standard Game" is available because of issues with implementing the setup for `MayWeekUtilitiesPlugin`, see below), and then relevant plugins are automatically enabled/disabled according to this. Then the setup API is called for all the enabled plugins that walk through the relevant config for each plugin. Mostly the API calls for the plugins just call (a useful subset of) the ask/answer functions of their config exports, though `TargetingPlugin` will first verify that it is safe to modify its config. To make the walking-through of config options clearer I added some additional explanatory text to various config options that `Setup Game` presents to the user.

**Limitations:** Because assigning the Umpire and Chief of Police requires players to have already been added, `Setup Game` requires players to have been added to the game (and will tell the user if they try to use it without players having been added). For this reason it does not run automatically on startup of a fresh install as I suggested in response to #140. Also, because several `MayWeekUtilitiesPlugin` config options depend on whether teams are enabled, I couldn't see a sensible way of walking through the config in the current framework, so I've parked that for now. (Perhaps we need some sort of ConfirmationDependency MetaComponent that can hide components depending on the response given to an initial`Checkbox` component?)

**Testing:** Manual.

[**Update:** In the latest commit I switched from only actually enabling/disabling the plugins after running through the config options to enabling/disabling the plugins before even checking whether any players have been added. This is so that `PlayerImporterPlugin` is guaranteed to be enabled when the user is told that they need to add players first. When making this change I took the opportunity to rework slightly how plugins are enabled/disabled, removing the unused `enable` and `disable` methods of `AbstractPlugin` and replacing them with an actually functional `@property` `enabled` that handles iteracting with the `GenericStateDatabase`.]

### DIFF
--- a/AU2/plugins/AbstractPlugin.py
+++ b/AU2/plugins/AbstractPlugin.py
@@ -131,7 +131,7 @@ class AbstractPlugin:
     @enabled.setter
     def enabled(self, val: bool):
         if not isinstance(val, bool):
-            raise TypeError(f"{self.__class__.__name__} must be a boolean, not '{type(val)}'")
+            raise TypeError(f"{self.__class__.__name__}.enabled must be a boolean, not '{type(val)}'")
         GENERIC_STATE_DATABASE.plugin_map[self.identifier] = val
 
     def process_all_events(self, _: List[Event]) -> List[HTMLComponent]:
@@ -214,12 +214,16 @@ class AbstractPlugin:
 
     def on_request_setup_game(self, game_type: str) -> List[HTMLComponent]:
         """
-        Walk through config options
+        Walk through config options. Generally this method should call the `ask` functions of important
+        `ConfigExport`s of this plugin and return the results concatenated together.
+        Make sure that all the components have unique identifiers.
         """
         return []
 
     def on_setup_game(self, htmlResponse) -> List[HTMLComponent]:
         """
-        Effect config options
+        Effect config options. Generally this method should call the `answer` functions of the same `ConfigExport`s
+        called by `on_request_setup_game` above, passing `htmlResponse` to each, then returning the results concatenated
+        together.
         """
         return []
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -727,10 +727,18 @@ class CorePlugin(AbstractPlugin):
     def answer_core_plugin_update_config(self, htmlResponse):
         enabled_plugins = htmlResponse[self.config_html_ids["Plugins"]]
         enabled_plugins.append("CorePlugin")
+        newly_enabled = []
+        newly_disabled = []
         for ident, p in AVAILABLE_PLUGINS.items():
-            p.enabled = (ident in enabled_plugins)
+            to_enable = (ident in enabled_plugins)
+            if to_enable and not p.enabled:
+                newly_enabled.append(ident)
+            if not to_enable and p.enabled:
+                newly_disabled.append(ident)
+            p.enabled = to_enable
         return [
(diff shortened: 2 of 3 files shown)


AU2/frontends/inquirer_cli.py
```diff
@@ -593,7 +594,7 @@ def render(html_component, dependency_context={}):
         a = inquirer_prompt_with_abort(q)
         return {html_component.identifier: a["q"] == "Yes"}
 
-    elif isinstance(html_component, HiddenTextbox):
+    elif isinstance(html_component, HiddenTextbox) or isinstance(html_component, HiddenJSON):
```
jsyiek: Let's create a parent class `HiddenComponent` that these two inherent from. Fine to be blank, just is a bit clearer imo.

AU2/plugins/AbstractPlugin.py
```diff
@@ -125,11 +124,15 @@ def _dump_state(self):
...
+    @property
+    def enabled(self) -> bool:
+        return GENERIC_STATE_DATABASE.plugin_map.get(self.identifier, True)
 
-    def disable(self):
-        self.enabled = False
+    @enabled.setter
+    def enabled(self, val: bool):
+        if not isinstance(val, bool):
+            raise TypeError(f"{self.__class__.__name__} must be a boolean, not '{type(val)}'")
```
jsyiek: `self.__class__.__name__` looks like a mistype to me?

AU2/plugins/CorePlugin.py
```diff
@@ -657,17 +714,26 @@ def answer_core_plugin_delete_event(self, html_response_args: Dict):
...
             )
         ]
 
+    def answer_core_plugin_update_config(self, htmlResponse):
+        enabled_plugins = htmlResponse[self.config_html_ids["Plugins"]]
+        enabled_plugins.append("CorePlugin")
+        for ident, p in AVAILABLE_PLUGINS.items():
+            p.enabled = (ident in enabled_plugins)
+        return [
+            Label("[CORE] Plugin change success!")
```
jsyiek: Report what the change was

AU2/plugins/custom_plugins/PolicePlugin.py
```diff
@@ -154,7 +163,7 @@ def ask_set_special_ranks(self):
     def answer_set_special_ranks(self, htmlResponse):
         self.gsdb_set("Umpires", htmlResponse[self.html_ids["Umpires"]])
         self.gsdb_set("Chief of Police", htmlResponse[self.html_ids["CoP"]])
-        return [Label("[POLICE] Success!")]
+        return [Label("[POLICE] Set Umpire(s) and Chief(s) of Police")]
```
jsyiek:
```suggestion
        return [Label("[POLICE] Successfully set Umpire(s) and Chief(s) of Police")]
```

jsyiek (approved):
seems sensible to me

AU2/plugins/AbstractPlugin.py
thenewtonium: Oh it's meant to have a `.enabled` after it
since it's meant to be referring to the property being set.

jsyiek (approved):

Issue #165, opened 2025-11-07
Long processing time for targeting graph


thenewtonium:
Whilest working on #164, I noticed an issue with #143, which is that for the current game it takes a good number of seconds to actually process the targeting graph, leading to an annoying delay in the `Event -> Create` and `Event -> Update` interfaces, since `WantedPlugin` now queries the current targeting graph from `TargetingPlugin` whenever either of these exports are used. This issue isn't present using the Mich 2024 database which is what I tested #143 on.

As a 'sticking plaster' I suggest that we should add a config option to `TargetingPlugin` to allow for this feature to be turned off if it causes annoying delays. (Perhaps this could even be done automatically by having it time the graph generation, with two "on" options in the config, one which allows it to be turned off automatically and one which doesn't).

I suspect in this game some of the issue may have come from certain events having too many kills in them, since in the screenshot Rhiannon sent on the Discord there are 4 instances of searches being aborted due to excessive depth. So we should probably look to implementing some 'chunking' of deaths when processing updates as per the TODO @NerdyNinja11 wrote in the gigabolt code[^1].

[^1]: ```# TODO Adjust targeting plugin so that it can handle events with many deaths, by retargeting in chunks```

thenewtonium:
I guess another way of avoiding the long processing time being an issue would be to have `TargetingPlugin` cache players' *last emailed* targets and treat these as their 'canonical' targets, then only reconstruct the targeting graph when sending out emails (or when using `Assassin -> Summary`, which could show both sets of targets when they differ).

PR #171, opened 2025-11-09
[165] [`TargetingPlugin`] Process kills in chunks

thenewtonium:
**Problem:** It was noted even back in #71 that having events with more than 3 kills causes problems with the targeting algorithm. More recently, as documented in #165, in the current game computing the targeting graph has become slow due to an event containing 4 deaths.

**Solution:** Have the targeting algorithm process deaths in chunks of 3, rather than processing all the deaths in an event together.

**Testing:** The existing tests in `test_TargetingPlugin.py` still pass, and I have added an additional test for speedup due to chunking which is passed by this PR but failed by the old `TargetingPlugin`.

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -329,12 +329,13 @@ class TargetingPlugin(AbstractPlugin):
             if not deaths:
                 continue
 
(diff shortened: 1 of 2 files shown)


jsyiek (approved):
Let's explicitly keep this number unconfigurable. I'm almost at the point where I think we should make dangerous config exports impossible to change after game start unless you enter the Konami code on the Home Screen or smth.... idk.

Issue #136, opened 2025-03-20
`IncorrectPseudonymFormatter` causes crash if `_` present in neighbouring text


thenewtonium:
Essentially, if you have a situation like `text_with_underscores [1] `, the `Suggestion` identifier has more than one `_` which makes `original, replacement = suggestion_str.split("_")` cause a crash.

At the very least we should catch this error.

To actually address the issue, perhaps something like the following: in the identifier for the suggestion, we escape all `_` as (e.g.) `\_` (after escaping all `\` as `\\`), then when later extracting `original` and `replacement` from the identifier, split over a regex pattern which ignores `\_`, and "de-escape" the two halves.

thenewtonium:
For a longer-term fix: one of the PRs I've been working on (for #45) introduces a `HiddenJSON` component which could perhaps be used to rewrite the sanity check framework in a way that doesn't require `_` shenanigans.

PR #173, opened 2025-11-14
[136] Simpler `SanityCheck` framework

thenewtonium:
**Problem:** The bug raised in #136 shows the awkwardness of encoding the effect of sanity check suggestions purely in an identifier string.

**Solution:** In this PR, the processing of sanity checks is simplified by making use of the `HiddenJSON` component introduced in #163 replace the myriad `HiddenTextbox` components previously being used to pass information between `ask_generate_pages` and `answer_generate_pages`. `Suggestion`s then have their `str`-valued `identifier` attribute replaced with a `dict`-valued `data` attribute, and these are what is passed to each `SanityCheck`'s `fix_event` method instead of suggestion identifiers. In particular, for`IncorrectPseudonymFormatter` and `PlayerNotDead` these `dict`s have keys `original` and `replacement` to specifiy the substitutions that should be made.

**Testing:** I have created unit tests for all three current `SanityCheck`s. This involved an expansion of `test_utils`. In particular, I added a `ProxyEvent` class to access `Event`s created by the testing framework. `ProxyAssassin.kills` now returns a `ProxyEvent` which required editing various existing tests. All but one of the unit tests pass; the only one that does not is `test_LocalBackupPlugin`, as the code it is testing was moved to `CorePlugin` in #135.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -1,6 +1,7 @@
 import glob
 import os.path
 
+from collections import defaultdict
 from typing import Any, Callable, Dict, List, Tuple
 
 from AU2 import BASE_WRITE_LOCATION
@@ -33,7 +34,7 @@ from AU2.plugins import CUSTOM_PLUGINS_DIR
 from AU2.plugins.AbstractPlugin import AbstractPlugin, Export, ConfigExport, HookedExport, DangerousConfigExport, \
     AttributePairTableRow
 from AU2.plugins.AvailablePlugins import __PluginMap
-from AU2.plugins.constants import COLLEGES, WATER_STATUSES
+from AU2.plugins.constants import COLLEGES, HEADLINE_TRUNCATION_CUTOFF, WATER_STATUSES
 from AU2.plugins.sanity_checks import SANITY_CHECKS
 from AU2.plugins.util.game import get_game_start, set_game_start, get_game_end, set_game_end
 from AU2.plugins.util.date_utils import get_now_dt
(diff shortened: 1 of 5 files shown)


AU2/plugins/CorePlugin.py
```diff
@@ -742,101 +743,68 @@ def answer_core_plugin_update_config(self, htmlResponse):
     def ask_generate_pages(self):
         components = []
         event_count = 0
+        suggestion_data: Dict[str, Dict[str, Dict[str, dict]]] = {}
```
jsyiek: We have a data class JSON library in this project. Not a hard requirement from me, but I think you could improve the readability of this type here with that. Not blocking.

AU2/plugins/CorePlugin.py
```diff
@@ -742,101 +743,68 @@ def answer_core_plugin_update_config(self, htmlResponse):
...
-                        HiddenTextbox(
-                            identifier=f"SanityCheck_{event_count}_{sanity_check_count}_{i}_explanation",
-                            default=suggestion.explanation
-                        )
-                    )
-                    explanations.append(suggestion.explanation)
-
-                sanity_check_count += 1
+                    s_id = f"{sanity_check_identifier}_{i}"
+                    suggestion_data.setdefault(e.identifier, {}).setdefault(sanity_check_identifier, {})[s_id] = \
```
jsyiek: `defaultdict` might be more ergonomic for your needs?

AU2/plugins/CorePlugin.py
```diff
@@ -742,101 +743,68 @@ def answer_core_plugin_update_config(self, htmlResponse):
...
+                    explanations.append((suggestion.explanation, s_id))
 
             if explanations:
+                c_id = f"SanityCheck_{event_count}_explanations"
                 components.append(
                     SelectorList(
-                        identifier=f"SanityCheck_{event_count}_explanations",
-                        title=f"[SANITY CHECK] {e.identifier}",
+                        identifier=c_id,
+                        title=f"[SANITY CHECK] {e.text_display()[:50]}",
```
jsyiek: Let's move magic constant `50` to a named variable.

Issue #142, opened 2025-03-23
'Quick Update' exports


thenewtonium:
This was discussed a while back but no export was actually made. The idea of a 'quick update' is to have an export which, rather than going through all the data for an `Assassin`/`Event` in sequence, presents a list of attributes that can be edited and then just displays the interface for that, avoiding the user having to mash <kbd>↵</kbd> to skip through the irrelevant parts of the `Update` exports. For dealing with attributes associated with particular plugins we can use an approach similar to #139, perhaps. 

The modifications to how `options_functions` is handled introduced by #83 also allow for a couple of enhancements i.e.:
- Not including certain options if they aren't applicable (e.g. if there are no participants don't ask about reports, if there are only police don't ask about competency, etc.)[^1]
- Showing a summary of the selected `Assassin`/`Event` before asking the user to choose which attribute they wish to modify[^2]

[^1]: This is possible because the return value of each function in `options_functions` is now passed to the next function
[^2]: This is possible because the functions in `options_functions` can now return arbitrary `HTMLComponent`s

jsyiek:
This should live in UIConfig

jsyiek:
Then, it could be implemented by using (probably new) tech that gathers all questions from an export and wraps them with a "Selector List"

PR #175, opened 2025-11-16
Dedicated `Assassin -> Pseudonyms` and `Event -> Reports` exports

thenewtonium:
**Problem:** In #142 and previously we discussed the idea of 'Quick Update' exports, to speed up the process of updating a single attribute of an Assassin/Event. However, the `Dependency`s in `Event -> Update` seems to cause some headaches in creating a generic 'Quick Update' framework as suggested in #142 .

**Solution:** This PR introduces `Assassin -> Pseudonyms` and `Event -> Reports` exports that only update the respective attributes, as adding pseudonyms and reports seem to be most common changes that an umpire would want to make to existing assassins and events.

The `ask` and `answer` functions for these exports are also called by `Assassin -> Update` resp. `Event -> Create`/`Event -> Update` to generate the relevant components and action the responses.

**Testing:** Manual.


AU2/plugins/CorePlugin.py
```diff
@@ -200,6 +208,13 @@ def __init__(self):
                 self.answer_core_plugin_update_event,
                 (self.gather_events,)
             ),
+            Export(
+                "core_event_update_reports",
+                "Event -> Reports",
```
jsyiek:
```suggestion
                "Event -> Update Reports",
```

AU2/plugins/CorePlugin.py
```diff
@@ -168,6 +169,13 @@ def __init__(self):
                 self.answer_core_plugin_update_assassin,
                 ((lambda: ASSASSINS_DATABASE.get_identifiers()),)
             ),
+            Export(
+                "core_assassin_update_pseudonyms",
+                "Assassin -> Pseudonyms",
+                self.ask_core_plugin_update_pseudonyms,
+                self.answer_core_plugin_update_pseudonyms,
+                ((lambda: ASSASSINS_DATABASE.get_identifiers()),)
```
jsyiek:
```suggestion
                (ASSASSINS_DATABASE.get_identifiers,)
```

AU2/plugins/CorePlugin.py
```diff
@@ -651,6 +659,28 @@ def answer_core_plugin_update_assassin(self, html_response_args: Dict):
             return_components += p.on_assassin_update(assassin, html_response_args)
         return return_components
 
+    def ask_core_plugin_update_pseudonyms(self, arg: str):
+        assassin = ASSASSINS_DATABASE.get(arg)
+        return [
```
jsyiek: `None` check required?

jsyiek (approved):

PR #176, opened 2025-11-16
[144] Config option for escaping HTML

### LINKED ISSUE #144: HTML escaping should be disabled by default and enabled via a config option
I don't trust future players who look at AU2 to listen to my warning.

```
def soft_escape(string: str) -> str:
    """
    Escapes html and adds <br /> to newlines only if not prefixed by HTML_REPORT_PREFIX
    """

    # umpires may regret allowing this
    # supposing you are a clever player who has found this and the umpire does not know...
    # please spare the umpire any headaches
    # and remember that code injection without explicit consent is illegal (CMA sxn 2/3)
    if not string.startswith(HTML_REPORT_PREFIX):
``

### PR DESCRIPTION
**Pull request by thenewtonium**
This is an implementation of #144. The config option also appears in `Setup Game` which is also a way of informing the user of the necessity of prefixing player reports / pseudonyms containing HTML with `<!--HTML-->`.

### DIFF
--- a/AU2/plugins/util/game.py
+++ b/AU2/plugins/util/game.py
@@ -12,16 +12,17 @@ HTML_REPORT_PREFIX = "<!--HTML-->"
 
 def soft_escape(string: str) -> str:
     """
-    Escapes html and adds <br /> to newlines unless prefixed by HTML_REPORT_PREFIX and html is enabled
+    Escapes html and adds <br /> to newlines unless prefixed by HTML_REPORT_PREFIX *and* html is enabled using the
+    config option CorePlugin -> Allow/disallow HTML.
     """
 
     # umpires may regret allowing this
     # supposing you are a clever player who has found this and the umpire does not know...
     # please spare the umpire any headaches
     # and remember that code injection without explicit consent is illegal (CMA sxn 2/3)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: AU2/plugins/util/game.py [hunk 1]
```diff
@@ -12,18 +12,32 @@
...
     """
 
     # umpires may regret allowing this
     # supposing you are a clever player who has found this and the umpire does not know...
     # please spare the umpire any headaches
     # and remember that code injection without explicit consent is illegal (CMA sxn 2/3)
-    if not string.startswith(HTML_REPORT_PREFIX):
+    if not (get_allow_html() and string.startswith(HTML_REPORT_PREFIX)):
         return escape(string).replace("\n", "<br />\n")
     return string
```
jsyiek: seems clearer to me to invert the condition
```suggestion
    if get_allow_html() and string.startswith(HTML_REPORT_PREFIX):
        return string
    return escape(string).replace("\n", "<br />\n")
```

jsyiek (approved):
fine to merge with/without my suggestions, no further review needed here

thenewtonium:
Ok that's true. I'll merge it just before creating the next release, then, to make sure we don't forget to tell them.

Issue #155, opened 2025-10-29
Standardised plugin state framework


thenewtonium:
`PolicePlugin` includes a framework for storing and retrieving config options, with default values for each. A similar framework has also been implemented in other plugins. It might be sensible to include such a framework in the `AbstractPlugin` class to be used by all these plugins rather than in being re-implemented from scratch in every plugin.

Issue #119, opened 2025-03-06
Local config database


thenewtonium:
AU2 has various UI settings (see #56, #35, #101), but since these are stored in `GenericStateDatabase` they get synced with SRCF, which is somewhat annoying. I propose creating a new database file (`LocalConfigDatabase.json` maybe), which doesn't get uploaded to SRCF, to store these options.

PR #181, opened 2025-11-19
[119] Local Config Database

thenewtonium:
**Problem:** As described in #119, there are various config settings (`CorePlugin -> Hide menu options`, `CorePlugin -> Pin menu options`, and `UI Config -> Toggle UI options`) which only affect the UI and so should probably not be synced with SRCF. 

**Solution:** This PR adds a new database, `__LocalConfigDatabase.json`, that is not uploaded to SRCF or backed up by `LocalBackupPlugin`, to store these UI settings. (The `__` in the filename was inspired by its use for private attributes in Python).

In light of #155, I have also implemented a standardised plugin state framework for the local config database. This takes the form of a function `local_config_property` that produces `property` objects whose getters and setters interact with the LocalConfigDatabase.

**Testing:** Manual. ~~For testing the changes to SRCFPlugin I have only tested that `__LocalConfigDatabase.json` is excluded from the *downloaded* files (by uploading `__LocalConfigDatabase.json` manually), as I did not want to interfere with the current live game by testing an upload.~~ UPDATE: I've realised that I can test this by temporarily changing `SRCF_WEBSITE` to `localhost` and running a local ssh server, which I have now done.

**Update:** This PR furthermore introduces the concept of 'local' plugins, whose enabled/disabled state is stored in `LocalConfigDatabase`, rather than `GenericStateDatabase` and so not synced with SRCF. These are `LocalBackupPlugin`, `PlayerImporterPlugin, `RandomGamePlugin`, `SRCFPlugin`, and `UIConfigPlugin`.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -984,7 +984,7 @@ class CorePlugin(AbstractPlugin):
             ),
             Checkbox(
                 self.identifier + "_include_local",
-                "Also reset UI config?",
+                "Also reset LOCAL config?",
                 False
             ),
         ]


AU2/database/LocalConfigDatabase.py
```diff
@@ -0,0 +1,30 @@
...
+@dataclass_json
+@dataclass
+class LocalConfigDatabase(PersistentFile):
+    """
+    This class is for storing local settings that shouldn't be uploaded to SRCF or saved in backups
+    """
+    arb_state: Dict[str, Any] = field(default_factory=dict)
+
+    # __ signals that the database file shouldn't be uploaded
+    WRITE_LOCATION = os.path.join(BASE_WRITE_LOCATION, "__LocalConfigDatabase.json")
```
jsyiek: Make `__` a constant somewhere please

PR #182, opened 2025-11-20
`DangerousConfigExport`s: conditions and explanations for 'danger'

thenewtonium:
**Problem:** Currently the reasons why each `DangerousConfigExport` is 'dangerous' is not well communicated. Also, certain DCEs are only 'dangerous' under specific conditions, but they currently marked as always 'dangerous'.

**Solution:** `DangerousConfigExport`s now have a `danger_explanation` attribute which is a function returning a string describing the danger, or `""` if the `Export` is not currently dangerous. In particular
- the targeting config options are only marked as 'dangerous' after targets have been sent out
- the auto-competency setting is only marked as dangerous if autocompetency is enabled, as switching from manual to auto is fine

`CorePlugin -> Set game start` is now no longer a `DangerousConfigExport`, since it is unlikely to be changed recklessly and the explanation in the export itself added in #163 seems sufficient to make umpires aware of the effects.

**Testing:** Manual.

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -169,8 +169,8 @@ class TargetingPlugin(AbstractPlugin):
 
     def danger_explanation(self) -> str:
         if int(GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get("last_emailed_event", -1)) > -1:
-            return "Targets have already been sent out to players. " \
-                   "Changing targeting settings will change players' targets!"
+            return ("Targets have already been sent out to players. "
+                   "Changing targeting settings will change players' targets!")
         else:
             ""
 


AU2/plugins/custom_plugins/TargetingPlugin.py
```diff
@@ -164,6 +167,13 @@ def on_data_hook(self, hook: str, data):
             max_event = data.get("secret_id", 100000000000000001) - 1  # - 1 needed to not include the current event
             data["targeting_graph"] = self.compute_targets([], max_event)
 
+    def danger_explanation(self) -> str:
+        if int(GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get("last_emailed_event", -1)) > -1:
+            return "Targets have already been sent out to players. " \
+                   "Changing targeting settings will change players' targets!"
```
jsyiek: prefer implicit continuations

```suggestion
            return ("Targets have already been sent out to players. "
                   "Changing targeting settings will change players' targets!")
```

Issue #95, opened 2024-11-30
Modify colours of dead player's real names slightly


NerdyNinja11:
Currently, when a major player dies, their real name can be lost in a sea of identically-coloured pseudonyms.

Changing the colour slightly (maybe darkening it) would make this stand out more on the website.

PR #190, opened 2025-11-28
[95] Darker real names + config option to set real name brightness

thenewtonium:
**Problem:** As Peter pointed out in #95, it can be hard to make out the real name of a dead player with many pseudonyms.

**Solution:** As suggested, this PR darkens real names slightly compared to pseudonyms. However, it also adds a config option to set this relative brightness; in particular the umpire can set the brightness to 1 for AU2's current behaviour. I've set  the default brightness to be `0.7`, which looks like this:

Bildschirmfoto 2025-11-28 um 6 14 24 PM

**Testing:** The function that adjusts the brightness of colours has unit tests in `test_render_utils.py`. The rest of the testing was manual.


AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -39,6 +42,33 @@ def __init__(self):
 
         self.exports = []
 
+        self.config_exports = [
+            ConfigExport(
+                "page_generator_plugin_real_name_brightness",
+                "PageGeneratorPlugin -> Set real name brightness",
```
jsyiek: I don't think this name is particularly helpful. How about "Adjust name brightness in news headline" or similar?

AU2/plugins/util/render_utils.py
```diff
@@ -150,6 +151,44 @@ def get_color(pseudonym: str,
...
+        str: the hexcode of the brightness-adjusted colour, (including an initial #).
+    """
+    rgb = (int(hexcode[i : i+2], 16) for i in (1, 3, 5))
+    factor = max(factor, 0)
+    scaled = (int(factor * x) for x in rgb)
+    capped = (min(255, x) for x in scaled)
+    return '#' + "".join(f"{x:02x}" for x in capped)
+
+
+DEFAULT_REAL_NAME_BRIGHTNESS = 0.7
```
jsyiek: Let's put global constants at top of file

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
thenewtonium: I think it the name should make clear that it's specifically for real names, otherwise sure.

AU2/plugins/custom_plugins/PageGeneratorPlugin.py
```diff
@@ -39,6 +42,33 @@ def __init__(self):
...
+                "page_generator_plugin_real_name_brightness",
+                "PageGeneratorPlugin -> Set real name brightness",
+                self.ask_set_real_name_brightness,
+                self.answer_set_real_name_brightness
+            ),
+        ]
+
+    def ask_set_real_name_brightness(self) -> List[HTMLComponent]:
+        return [
+            Label("Below you can set the brightness of real names relative to pseudonyms on news pages."),
```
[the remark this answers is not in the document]
thenewtonium: Actually there isn't a rule that it has to be <= 1, it's just that values > 1 make the real names *brighter* than pseudonyms which tends to be hard to read. E.g. this is what 1.3 relative brightness looks like
1 3-brightness-example
I guess this fact should be added to the explanatory text.

Wrt it being relative rather than absolute brightness: setting absolute brightness does make much sense to me seeing as the real names will be contrasting with different pseudonym colours that have different brightnesses, and the different pseudonym colours having different brightnesses also means that real names couldn't be set to be the same colours as pseudonyms.

And tbh I'm not convinced that it is any easier to visualise what say 50% absolute brightness will look like compared to 50% relative brightness. (I think expressing the brightness on a scale of 1-100 rather than is a good idea though!) My assumption is that umpires would either use the default setting, or set the real names to either be the same as pseudonyms or black, or if they really don't like either of those options they'd just iteratively check against the generated news pages.

Anyway this is what setting absolute brightness would look like with 30% brightness
0 3-abs-brightness

and with 70% brightness
0 7-abs-brightness

(Specifically, for these examples I took absolute brightness to be the sum of the r g and b values and then scaled the rgb of colours so that the total brightness by this definition matches (approximately) the specified percentage of 3 * 255 (the brightness of white by this definition))

PR #191, opened 2025-11-29
New targeting algorithm respecting teams / mentoring

### LINKED ISSUE #216: [`TargetingPlugin`] Incorrect warning when triangles in updated targeting graph
To record here a discussion on the Discord: Joel has pointed out that AU2 will give the warning `[TARGETING] WARNING: Two assassins target each other due to an unavoidable graph collapse` when the actual issue is having a triangle. I don't think this can be fixed (other than in a convoluted way) without possibly changing computed targets (hence why I've tagged this issue `not when live`).

thenewtonium:
Jamie and I were discussing prior to the current game having the targeting algorithm respect teams, since the umpires wanted to have a mentoring system, but didn't manage to implement this in time for the current game. This is still an interesting idea though, so here's an implementation of that.

The old targeting algorithm is still used when no teams are set, because I think it is better at dealing with seeds.

This PR supersedes #171 as a solution to #165 as it implements chunking of kills when updating the graph (for both the old and new targeting algorithms, which use the same function to process updates), but I'm going to keep #171 open because that specific feature is more urgent than this.

**Testing:** I've added a few unit tests to `test_TargetingPlugin.py`, two of which deal with checking that team constraints are respected, and one of which is copied over from #171. The teams editor UI was tested manually.

**Update:** This PR now also resolves #216.

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -829,6 +829,24 @@ class TargetingPlugin(AbstractPlugin):
                     player_teams,
                     allow_mutual_team_targets=True,
                     allow_mutual_seed_targets=True,
+                    allow_triangles=True
+                )
+
+                if success:
+                    response.append(
+                        Label("[TARGETING] WARNING: The targeting graph contains a triangle due to an unavoidable graph collapse."))
+                    continue
+
+                success = self.update_graph(
+                    response,
+                    targeting_graph,
+                    targeters_graph,
+                    deaths,
+                    player_seeds_set,
+                    player_teams,
+                    allow_mutual_team_targets=True,
+                    allow_mutual_seed_targets=True,
+                    allow_triangles=True,
                     allow_mutual_targets=True
                 )
 
@@ -854,6 +872,7 @@ class TargetingPlugin(AbstractPlugin):
             player_teams: Sequence[Set[str]] = (),
             allow_mutual_team_targets: bool = False,
             allow_mutual_seed_targets: bool = False,
+            allow_triangles: bool = False,
             allow_mutual_targets: bool = False,
             limit_checks=1000000
     ) -> bool:
@@ -931,8 +950,8 @@ class TargetingPlugin(AbstractPlugin):
             # (a triangle is any targeting of players (eg) Vendetta, OHare, and O-Ren Ishii such that
             # Vendetta targets OHare targets O-Ren Ishii targets Vendetta
             # V -> OH -> OR -> V)
-            # Ignore conditions: allow_mutual_targets=True
-            if not allow_mutual_targets:
+            # Ignore conditions: allow_triangles=True
+            if not allow_triangles:
                 # precondition: assume triangle elimination has been performed up to this point.
                 # While triangle elimination originated with the previous AU, Peter has a good description of it:
                 # "only check the targets of the newly assigned targets' targets,


AU2/plugins/custom_plugins/TargetingPlugin.py
```diff
@@ -232,8 +253,189 @@ def render_assassin_summary(self, assassin: Assassin) -> List[AttributePairTable
...
+        teams = [set(t) for t in GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get("teams", [])]
+        # if teams are not set, use old targeting algo instead
+        if not teams:
+            return self.old_compute_targets(response, max_event)
+
+        # reset the seed for determinism
+        random.seed(self.seed)
+
+        # collect all targetable assassins
+        players = [a for (a, model) in ASSASSINS_DATABASE.assassins.items() if not model.is_city_watch]
```
jsyiek: Let's use the new function from #205

AU2/plugins/custom_plugins/TargetingPlugin.py
```diff
@@ -232,8 +253,189 @@ def render_assassin_summary(self, assassin: Assassin) -> List[AttributePairTable
...
+
+            # shuffle list of players
+            remaining_players = list(players)
+            random.shuffle(remaining_players)
+
+            chain = []
+
+            def check_constraints(next_p: str, curr_p: str, prev_p: str,
+                                  no_triangles=True, no_teammates=True, no_mutuals=True) -> bool:
+                return not (
```
jsyiek: This is a very long condition. You need to break it down so this is easy to understand for people in future. For example, someone looking at this in future might not understand what `next_p`, `curr_p`, and `prev_p` mean. 

I would give this function its own doc string for the latter.

How you might break it down could be something along these lines:
```
no_mutual_targets = next_p not in target_graph.get(...)
no_teammate_targets = ...;

return no_mutual_targets and no_teammate_targets
``

File: AU2/plugins/custom_plugins/TargetingPlugin.py [hunk 3]
```diff
@@ -232,8 +253,189 @@ def render_assassin_summary(self, assassin: Assassin) -> List[AttributePairTable
...
+        player_seeds = {p for p in GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get("seeds", [])}
+        use_seeds_for_updates_only = GENERIC_STATE_DATABASE.arb_state.get(self.identifier, {}).get(
+            "use_seeds_for_updates_only", False)
+        if not use_seeds_for_updates_only:
+            teams.append(player_seeds)
+
+        # player -> teams mapping for checking seeding constraints
+        player_teammate_mapping = {p: {teammate for t in teams for teammate in t if p in t} for p in players}
+
+        def new_chain(targeting_graph: Dict[str, Set[str]],
```
jsyiek: This is getting complicated enough that I'm wondering if Chain should just be its own class to reduce the size of this function. We have a `new` function that creates a Chain, and then several functions that do operations on it. That feels like it should be a class for organization purposes, no?

AU2/plugins/custom_plugins/TargetingPlugin.py
```diff
@@ -232,8 +253,189 @@ def render_assassin_summary(self, assassin: Assassin) -> List[AttributePairTable
...
+
+            # check constraints for endpoints of chain (since actually a cycle)
+            if not (
+                    check_constraints(chain[0], chain[-1], chain[-2])
+                    and check_constraints(chain[1], chain[0], chain[-2])
+            ):
+                raise FailedToCreateChainException()
+
+            # save the new edges
+            for (targeter, targetee) in zip(chain, chain[1:] + [chain[0]]):
```
jsyiek: We're dealing with a time-consuming algorithm, so let's not over-allocate more than we need to. 

We can turn to C-bindings if execution time becomes an issue.

```suggestion
            for (targeter, targetee) in zip(chain, itertools.chain(chain[1:], [chain[0]])):
```

AU2/plugins/custom_plugins/TargetingPlugin.py
jsyiek: Let's have a `DangerousConfigExport` to swap between the two targeting graph implementations.

AU2/test/plugins/util/custom_plugins/test_TargetingPlugin.py
```diff
@@ -10,6 +13,21 @@ def valid_targets(num_players, targets):
...
+            for target in targets.get(member, ()):
+                assert target not in team
+    return True
+
+
+def seeds_respected(targets: Dict[str, List[str]], seeds: List[str]) -> bool:
+    for seed1 in seeds:
+        for seed2 in seeds:
+            assert seed2 not in targets[seed1]
+    return True
```
jsyiek: boolean return value seems pointless to me?

```suggestion
def seeds_respected(targets: Dict[str, List[str]], seeds: List[str]):
    for (seed1, seed2) in itertools.combinations(seeds, 2):
        assert seed1 not in targets[seed2]
        assert seed2 not in targets[seed1]
```

AU2/test/plugins/util/custom_plugins/test_TargetingPlugin.py
jsyiek: Can you add a test that kills a player and checks incrementally that the targeting graph update works if we don't have one?

jsyiek (changes requested):

thenewtonium:
It did spot some issues but also suggested some things I think we can ignore.

PR #195, opened 2025-12-03
Include wantedness status in emails

thenewtonium:
A PR to include wantedness status in emails and to have changes in wantedness status count towards 'updates' when selecting `UPDATES ONLY`. This is perhaps less necessary than #170 since crimes are entered manually by the umpire and thus will presumably be communicated to players anyway?

### DIFF
--- a/AU2/plugins/custom_plugins/WantedPlugin.py
+++ b/AU2/plugins/custom_plugins/WantedPlugin.py
@@ -295,8 +295,8 @@ class WantedPlugin(AbstractPlugin):
             for e in events:
                 wanted_manager.add_event(e)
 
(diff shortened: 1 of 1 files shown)


AU2/plugins/custom_plugins/WantedPlugin.py
```diff
@@ -82,6 +83,8 @@ def __init__(self):
             "Wanted": self.identifier + "_wanted"
         }
 
+        Assassin.__last_emailed_crime = self.assassin_property("last_emailed_crime", None, store_default=False)
```
jsyiek: Probably shouldn't be private if it's being written to by another class?

jsyiek (approved):
Fix city_watch/police thing and fine to merge

Issue #133, opened 2025-03-20
Duel page


thenewtonium:
Sometimes the duel takes place a while after the end of the game. In these circumstances, it makes sense to put the duel on its own news page, à la [Mich 2023](http://assassins.soc.srcf.net/archive/2023-mich/duel.html).

#132 introduces the concept of 'game end' into `CorePlugin`, so AU2 would be able to use this to automatically detect the duel.

The question is — if the duel occurs in, say, week 6, should the duel event be rendered on both the week 6 page and a separate duel page?

PR #198, opened 2025-12-07
[174] Choice of bounty plugin in `Setup Game`

thenewtonium:
**Problem:** AU2 currently has two different bounty plugins: `BountyPlugin`, which displays bounties in table form à la the Wanted list, and `BountyNewsPlugin` which emulates the AU1 style of setting bounties by using news headlines as bounties, but `Setup Game` doesn't help to configure this aspect of AU2.

**Solution:** This PR adds a stage to `Setup Game` that provides information about the two bounty plugins and choose which to enable.

**Testing:** Manual.

**Remarks:** An alternative approach might be to merge the two bounty plugins together and have the choice of table vs events be a plugin-specific config option. Merging them is somewhat fiddly (and conflicts with #133 and #188...) so if this is preferable this PR can serve as a stop-gap for the time being.

### DIFF
--- a/AU2/plugins/CorePlugin.py
+++ b/AU2/plugins/CorePlugin.py
@@ -88,6 +88,7 @@ GAME_TYPE_PLUGIN_MAP = {
     #     "WantedPlugin": True,
     # }
 }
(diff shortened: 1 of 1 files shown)


jsyiek (approved):
love it

PR #199, opened 2025-12-11
Sortable stats table

thenewtonium:
Rather than having the umpire fix a specific ordering to the stats page, it would be cool if visitors to the website could choose what column to sort the table by, à la some tables on Wikipedia.

This PR implements this by making use of the [table-sort-js](https://github.com/kyle-wannacott/table-sort-js) library and applying the relevant classes to the stats table (and in the case of the 'Died at' column, including appropriate `data-sort` attribute values).

I thought that it would be sensible to incorporate the .js file into AU2 itself so that it and the archive are self-contained. Because the library is distributed under an MIT license I have prepended the necessary text to AU2's license.


AU2/plugins/custom_plugins/ScoringPlugin.py
```diff
@@ -285,33 +282,38 @@ def _generate_stats_page(self,
...
         rows = []
 
-        full_players.sort(key=lambda a: STATS_ORDERING_KEYS[stats_order](score_manager, a))
+        def player_rating(a):
+            """
+            The values used to 'rank' players.
+            Players are sorted by 'rating' which is time of death for those that died before the end of open season
+            and game end + score for those that survived open season. If this is a tie then ties are broken by score.
+            """
+            return -score_manager.get_rating(a), -score_manager.get_score(a)
```
jsyiek: I'm not a fan of unlabeled tuples... can we use a dataclass (or named tuple) please so that `value[0]` and `value[1]` are clearly named?

AU2/plugins/custom_plugins/html_templates/table-sort.js
jsyiek: Why does AU2 need to store this file? Seems to me like this could just live in SRCF and then we wouldn't need to maintain it / amend our license?

Feel free to add appropriate logging to the tool so that people have an example of the API for a `table-sort.js` (you could link to this as an example) but I don't think we need to bundle the code here.

thenewtonium: if we don't want to bundle the code maybe there could be an export to download `table-sort.js` from the original repo.

jsyiek (approved):

PR #200, opened 2025-12-11
Prettier datetime format on stats and city watch pages

thenewtonium:
**Problem:** #180 implements various improvements to the killgraph visualisation, including rendering the dates of kills in a shorter (and imo more aesthetic) form. The stats and (since #160) ~~police~~ city watch pages include links to death events, and, while brevity isn't so important in this case, the current format is not the most readable or aesthetic, in my opinion.

**Solution:** This PR changes the format of the death datetimes on the stats and police pages from YYYY-MM-DD HH:MM to DD Mmm, HH:MM. It also adds a semi-colon to separate multiple deaths on the police page to visually separate the different datetimes more clearly, and has the list of death datetimes now in brackets after an overall count of the *number* of deaths for that member of police.


AU2/plugins/util/date_utils.py
```diff
@@ -5,6 +5,8 @@
 
 DATETIME_FORMAT = "%Y-%m-%d %H:%M"
 
+PRETTY_DATETIME_FORMAT = "%d %b, %H:%M %p"
```
jsyiek: 24 hour time AND an AM/PM specifier...? Why?

thenewtonium: My reasoning was that it should be consistent with the format used on the news pages

Issue #202, opened 2025-12-14
[Emails] Send updated targets when targets change due to updating an event


thenewtonium:
As discussed in #166 (which has AU2 store the previously-emailed competency deadline for each player so that they will be included in `UPDATES ONLY` when their competency deadline is extended) the current method of storing the last updated event has the flaw that if the targeting graph changes due to updating a previous event (if a kill is annulled, for example) then the targeting graph in AU2 could become out of sync with the targets emailed to players. So it would make sense to store each player's last emailed targets and compare against these when determining which players need to be sent updates, in case the targeting graph reconstructed up to the last emailed event is wrong.

Going further, since emailed targets are really the 'canonical' targets, perhaps it would be more useful for the targeting info displayed by `AssassinDependentCrimeEntry` to be based off of emailed targets? Of course, if we're only storing last emailed targets then it would only make sense to have targeting info in `Event -> Create` and not `Event -> Update`... Either way, it may be useful for `Assassin -> Summary` to tell the umpire *both* sets of targets.

PR #203, opened 2025-12-17
[202] Store last emailed targets as basis for update emails (and targets displayed in `Event -> Create`)

thenewtonium:
As set out in #202, in #170 @NerdyNinja11 and I agreed that it would make sense for changes in targets to be detected by storing the last set of targets to be emailed out, rather than reconstructing what they 'should' have been, so that changes in targets are still detected if e.g. targets change due to a kill being annulled.

Beyond the simple change agreed on in #170, it also occurred to me that it would make more sense for umpires to reference the targets actually sent to players. So this PR additionally

- Has `Assassin -> Summary` list *both* a player's current targets *and* the targets last emailed to them; and
- Uses last emailed targets for the targets displayed by `AssassinDependentCrimeEntry`. As this only makes sense for players' *current* targets, licitness information is now only displayed when creating and not when updating events.

A side of effect of the latter is to make #165 less urgent as the targeting graph is no longer reconstructed each time `Event -> Create` or event `Event -> Update` is used.

Of course, if you're not sold on the idea of treating emailed targets as the 'canonical' targets then it would be easy to remove that from this PR...

This PR also fixes a bug where the last emailed event and last emailed competency deadlines aren't uploaded to SRCF because the databases in memory aren't saved to disk between these being saved in memory and the databases being uploaded. Previously this wasn't too much of a problem -- worst case someone receives a superfluous email! -- but if AU2 tells umpires kills are illicit because it is out of sync with the actual targets emailed people it would be very bad.

**Testing:** Manual, by running a local ssh server to stand in for SRCF, as with #170.

### DIFF
--- a/AU2/plugins/custom_plugins/TargetingPlugin.py
+++ b/AU2/plugins/custom_plugins/TargetingPlugin.py
@@ -40,9 +40,6 @@ class FailedToCreateChainException(Exception):
     pass
 
 
-TOGGLE_INFO_DISPLAY_NAME = "Targeting Graph -> Toggle display"
-
-
 def filter_to_targetable(idents: Iterable[str]) -> List[str]:
     """Filters an iterable of assassin identifiers to a list of only those that are involved in targeting,
     i.e. full players"""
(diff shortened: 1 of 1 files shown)


jsyiek:
Ah well, if it doesn't work, just go ahead and merge

thenewtonium:
I ended up removing the option to toggle showing targets because of merge conflicts btw (and it being redundant when targets aren't being re-computed each time)

Issue #104, opened 2025-02-09
`header.html` generator


thenewtonium:
Currently the umpires have to edit `header.html` at various points in the game. Since most of the changes already correspond to changes in game state anyway it would make sense for this to be generated by AU2.

We would need to decide how to deal with
- the Awards page, since this isn't part of AU2 at the moment (unless an Awards plugin is implemented prior to this...)
- the Stats page, because it's not (currently) generated by `Generate Pages`, but by its own export
- the week 0 page; currently "Bounties" links to this this, but now we actually have a separate bounties page. Maybe make including week 0 a config option in case an umpire wants to do this for some reason?
- the news page for the duel (c.f. [Mich 2023](http://assassins.soc.srcf.net/archive/2023-mich/duel.html)). Perhaps this should be combined into how we deal with the stats page: currently `Scoring -> Generate stats page` asks when open season ended. This could be changed into a config option which would used by `Generate pages` to decide whether to generate a stats page *and* to put all events occuring after the end of open season on a `duel.html` page.

Presumably this would be a new API call, since the pages that need to be linked to depend on what plugins are loaded.

jsyiek:
> the Stats page, because it's not (currently) generated by Generate Pages, but by its own export

solving this issue should change that

PR #204, opened 2026-01-27
Rename 'police' to 'city watch'

thenewtonium:
@axellerationduetogravity says that the Guild is planning to change the name of the 'police' to 'city watch', and has requested that the necessary adjustments be made to AU2.

The adjustments made in this PR are:

- The 'police' page is generated under `citywatch.html` rather than `police.html`
- The police page is headered by 'The City Watch' rather than 'Police list', and references to the 'police' are replaced by 'city watch'
- The default rank name for the Chief of Police is now 'Commander of the City Watch' (this is based on the rank of the leader of the City Watch of Ankh-Morpork in Discworld, which according to Axel is the basis for the new name).
- The text introducing the list of of corrupt police on the Wanted page now reads 'Those that were corrupted by the dark side of the city watch....' rather than 'Those that were corrupted by the dark side of the police force....' (I am aware that the 'dark side' reference makes less sense without the word 'force' but I was unable to think of anything witty to replace it wit).
- The kill-graph visualisation now disambiguates police from full players by appending '(City Watch)' rather than '(Police)' to their names.

~~This PR does *not* alter references to police in the UI of AU2, nor internally in the code.~~ _**Update:** now it does._

Also, the default ranks retain 'Constable' in their names, since the City Watch of Ankh-Morpork also has constables. And of course the navbar on the website will need to be updated manually, given issue #104 is yet to be resolved..


jsyiek:
I'm fine with casual or city watch. If the guild is nuking any mention of police, we ought not to allow ourselves to be the bearers of the resulting tech debt.

PR #205, opened 2026-02-05
[`TargetingPlugin`] Exclude City Watch from seeding

thenewtonium:
**Problem:** Currently it is possible to set city watch players as seeds for the targeting graph, but doing so causes a crash when computing targets.

**Solution:** This PR both removes the city watch players from the options when setting player seeds, and has `compute_targets` filter out any seeds that are city watch. The reason for doing both is that AU2 allows users to change a player between full player and city watch provided that targets are yet to be send out.


jsyiek:
> Wouldn't it make more sense for the filter to be 'can have targets' rather than specifically about seeding?

yes I'd say so

PR #210, opened 2026-02-07
AU2 Packager & Releaser

jsyiek:
Adds a GitHub workflow to release a new AU2 version, allowing us to release even when not at a computer and without having a development environment set up.


thenewtonium (approved):
Seems reasonable as far as I can tell.

PR #221, opened 2026-03-18
Fix crash when generating stats page with hidden deaths

thenewtonium:
This is the branch that [v1.6.3-pre1](https://github.com/jsyiek/AU2/releases/tag/v1.6.3-pre1) is based off.

**Problem:** Generating the stats page caused a crash in the current game. Reading the coredump sent by Joel, this seems to be triggered by trying to generate a url for hidden events.

**Solution:** Twofold: firstly, I have fixed the `event_url` function to not crash on hidden events, and secondly have `ScoringPlugin` not generate a link for hidden deaths.

**Testing:** I have added a unit test for `event_url` to prevent regression, and manualy tested that page generation is correct using the database for the current game.

### DIFF
--- a/AU2/plugins/util/render_utils.py
+++ b/AU2/plugins/util/render_utils.py
@@ -13,7 +13,7 @@ from AU2.plugins.util.CompetencyManager import CompetencyManager
 from AU2.plugins.util.DeathManager import DeathManager
 from AU2.plugins.util.WantedManager import WantedManager
 from AU2.plugins.constants import WEBPAGE_WRITE_LOCATION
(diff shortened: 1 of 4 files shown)


jsyiek:
>  Reading the coredump sent by Joel

🥹

AU2/plugins/util/render_utils.py
jsyiek: This function returns an optional in everything except name. Let's change its return type to `Optional` and return `None` instead of `""`.

jsyiek (approved):
Thanks for putting out the fix!!

PR #228, opened 2026-03-29
[227/222] KillEntry supporting thunderbolting

thenewtonium:
**Problem:** This PR solves two problems together: firstly, that the KillEntry UI gets unwieldy with many participants, and secondly that the current method of thunderbolting (having the umpire make a kill) turned out to be problematic in #180: hiding death events means they can't be linked to on the stats page, and having a flag for the gigabolt is problematic because it's invisible to the user.

**Solution:** For the first problem, this PR changes `AssassinDependentKillEntry` to have multiple stages, with the user first selecting which players died, and then selecting who killed them. 
I have made it so that in an event each player can only have one killer, since having multiple killers doesn't play well in scoring, although it's worth having a discussion whether it's better to leave the option of multiple kills and formalise how they're dealt with.

For the second problem, the updated kill entry UI gives an option for each death to by thunderbolt, ~~which is implemented in the backend as a player killing themselves.~~
~~Originally I implemented it as having thunderbolts be kills of the form `("", victim)`, but I realised that having thunderbolts be self-kills was more or less compatible with the existing handling of kills, with the only modfifications required being of how kills and conkers are counted for thunderbolted players, but this is minor since the scores of dead players don't really matter. I also updated the May Week scoring to account for thunderbolts, though we wouldn't expect thunderbolting to be a thing in May Week.~~ [see update below]

The gigabolt is updated to use these thunderbolt kills, and hence I took the opportunity to resolve #222 (merging gigabolt deaths into one event) in this PR.

**Testing:** This PR includes several unit tests to verify the ability of the various Managers + the targeting algorithm to cope with thunderbolt deaths, (which I had created when thunderbolts were going to be encoded as `("", victim)`, arguably they are less necessary with self-kills, but we might as well have them). The UI I tested manually.

**Update:** Further to discussions below, kills are now stored as `NamedTuple`s and thunderbolts are implemented with a `None` value for the 'killer'.

### DIFF
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -356,7 +356,7 @@ def render(html_component, dependency_context={}):
         q = [inquirer.List(
             name=victim,
             message=f"Who killed {escape_format_braces(victim)}?",
-            choices=[a for a in assassins if a != victim] + [("(Thunderbolt)", "")],
+            choices=[a for a in assassins if a != victim],
             default=default_killers[victim],
         ) for victim in deaths]
         victim_killer_mapping = inquirer_prompt_with_abort(q)
@@ -409,8 +409,6 @@ def render(html_component, dependency_context={}):
         mapping = {}
         defaults = []
         for (a1, a2) in kills:
-            if html_component.ignore_thunderbolts and not a1:
-                continue
             key = f"{a1} kills {a2}"
             mapping[key] = (a1, a2)
             if (a1, a2) in html_component.default:
--- a/AU2/frontends/inquirer_cli.py
+++ b/AU2/frontends/inquirer_cli.py
@@ -356,8 +356,10 @@ def render(html_component, dependency_context={}):
         q = [inquirer.List(
             name=victim,
             message=f"Who killed {escape_format_braces(victim)}?",
-            choices=[a for a in assassins if a != victim],
-            default=default_killers[victim],
+            # note: we use 'self-kills' for thunderbolts, for backwards compatibility.
+            # so long as thunderbolt events aren't edited on an older version of AU2 it will be able to cope with this
+            choices=[a for a in assassins if a != victim] + [("(Thunderbolt)", victim)],
+            default=default_killers.get(victim),
         ) for victim in deaths]
         victim_killer_mapping = inquirer_prompt_with_abort(q)
         return {html_component.identifier: [(killer, victim) for victim, killer in victim_killer_mapping.items()]}
@@ -409,6 +411,8 @@ def render(html_component, dependency_context={}):
         mapping = {}
         defaults = []
         for (a1, a2) in kills:
+            if html_component.exclude_thunderbolts and a1 == a2:
+                continue
             key = f"{a1} kills {a2}"
             mapping[key] = (a1, a2)
             if (a1, a2) in html_component.default:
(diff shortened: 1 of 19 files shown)


AU2/frontends/inquirer_cli.py
```diff
@@ -345,25 +345,24 @@ def _render_report_editor(old_report: Report) -> Optional[Report]:
...
-        # TODO: Confirm what happens if option in default isn't in choices
-        a = inquirer_prompt_with_abort(q)["q"]
-        a = tuple(potential_kills[k] for k in a)
-        return {html_component.identifier: a}
+        deaths = set(inquirer_prompt_with_abort(q)["q"])
+        default_killers = {victim: killer for killer, victim in html_component.default if victim in deaths}
+        q = [inquirer.List(
+            name=victim,
+            message=f"Who killed {escape_format_braces(victim)}?",
+            # note: we use 'self-kills' for thunderbolts, for backwards compatibility.
```
jsyiek: TBH I'm not thrilled with the self-kills quirk. This isn't what's happening and it creates all sorts of potential edge cases for future modules and game post-processing (e.g., stats) that we may not be seeing now.

The advantage of the previous system is that it had an implicitly defined nullable -- the empty string. What I'd like to see is a similar system, maybe with a more explicit nullable. 

I.e., a death event looks like this:
```
class DeathEvent:
    killer: Optional[Assassin]
    victim: Assassin
```

I don't suggest you return this class from `inquirer_cli`, but I'd like to see a design that captures the essence of that `Optional` in a way that doesn't use a real `Assassin`. I would be OK making a fundamental schema change to the `Event` to capture this -- the model for `Event` (and the entire AU2) was made with Don O'Hare's may week as the focus so there are rough edges like this that were definitely the wrong decision by me in retrospect. Hindsight is 20/20!

What do you think? What would be the lift required to make that change?

AU2/test/plugins/util/custom_plugins/test_CityWatchPlugin.py
```diff
@@ -71,3 +74,31 @@ def test_resurrect_as_city_watch(self):
...
+            |--------> 1
+        """
+        p = some_players(20)
+        game = MockGame().having_assassins(p)
+        game.assassin(p[0]).with_accomplices(*p[1:10]).are_city_watch()
+
+        game.assassin(p[0]).kills(p[10]).then()\
+            .assassin(p[0]).kills(p[11]).then()\
+            .assassin(p[12]).is_thunderbolted().then()\
+            .assassin(p[1]).is_thunderbolted()
```
jsyiek: nitpick

```suggestion
        game.assassin(p[0]).kills(p[10])\
            .then().assassin(p[0]).kills(p[11])\
            .then().assassin(p[12]).is_thunderbolted()\
            .then().assassin(p[1]).is_thunderbolted()
```

AU2/frontends/inquirer_cli.py
thenewtonium: Having an explicit nullable wouldn't take much work; the `("", victim)` system is already in the branch and can just be reverted to, and to make the nullable explicit the `""` could just be replaced by `None`. All the modifications I made to deal with an empty string can already deal with `None`.

It doesn't seem worth it to be to change deaths to a specific class, apart from maybe a `NamedTuple` which would allow it to be still treated as a tuple? `dataclass_json` doesn't support `NamedTuple`s properly natively, so the `Event.kills` field would need a custom decoder, but I think that's fine.

